PR #464, opened 2025-10-28
Reworked table processing. Added tracks information save/load

vstempen:
The change moves table processing by Sqlite into Cpp code with placing tables data into memory. It creates 3 separate table processor instances for event table browsing, table search and sample table. It mimics sqlite filtering to some point. The change also implemented track information and histogram save/load.


tomk-amd:
Implemented in dev branch instead

PR #545, opened 2025-12-03
Add Main Thread Indication 

Sajeeth-Wimalasuriyan:


src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -67,6 +67,29 @@ FlameTrackItem::FlameTrackItem(DataProvider&                      dp,
...
+            {
+                m_show_pill_label = true;
+                m_pill_label      = "MAIN";
+            }
+        }
+    }
+    else
+    {
+        spdlog::error("TrackItem: Invalid track id {}", m_id);
+        return;
```
tomk-amd: Don't return, let the constructor finish

Sajeeth-Wimalasuriyan: done

PR #569, opened 2025-12-12
Loading UI Redesign 

Sajeeth-Wimalasuriyan:


src/view/src/widgets/rocprofvis_widget.cpp
```diff
@@ -961,5 +962,61 @@ EditableTextField::SetOnTextCommit(const std::function<void(const std::string&)>
...
+{
+    ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(20, 16));
+    ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, ImVec2(10, 12));
+    ImGui::PushStyleVar(ImGuiStyleVar_WindowRounding, 10.0f);
+    ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(12, 6));
+    ImGui::PushStyleVar(ImGuiStyleVar_FrameRounding, 6.0f);
+    ImGui::PushStyleVar(ImGuiStyleVar_ChildRounding, 6.0f);
+
+    m_style_var_count = 6;
+}
```
tomk-amd: Why set new values here from magic numbers?  Can we use the values from settings?

PR #585, opened 2025-12-31
Vs add argument data

vstempen:
## Motivation

New feature to fetch event arguments. 

## Technical Details

Created new storage class based on ExtData class to store additional argument properties.
Store argument object into the same array as extended data
Those two objects can be distinguished by category name or enumeration value

### DIFF
--- a/src/controller/inc/rocprofvis_controller_enums.h
+++ b/src/controller/inc/rocprofvis_controller_enums.h
@@ -1036,14 +1036,6 @@ typedef enum rocprofvis_event_data_category_enum_t
 
     // Event argument data
     kRocProfVisEventArgumentData,
-    // Event argument position
-    kRocProfVisEventArgumentDataPosition,
-    // Event argument type
-    kRocProfVisEventArgumentDataType,
-    // Event argument name
-    kRocProfVisEventArgumentDataName,
-    // Event argument value
-    kRocProfVisEventArgumentDataValue,
 
 } rocprofvis_event_data_category_enum_t;
 


src/model/src/database/rocprofvis_db_profile.cpp
```diff
@@ -350,6 +350,34 @@ int ProfileDatabase::CallbackAddExtInfo(void* data, int argc, sqlite3_stmt* stmt
...
+int ProfileDatabase::CallbackAddArgumentsInfo(void* data, int argc, sqlite3_stmt* stmt, char** azColName) {
+    ROCPROFVIS_ASSERT_MSG_RETURN(data, ERROR_SQL_QUERY_PARAMETERS_CANNOT_BE_NULL, 1);
+    void*  func = (void*)&CallbackAddArgumentsInfo;
+    rocprofvis_db_sqlite_callback_parameters* callback_params = (rocprofvis_db_sqlite_callback_parameters*)data;
+    ROCPROFVIS_ASSERT_MSG_RETURN(callback_params->db_instance != nullptr, ERROR_NODE_KEY_CANNOT_BE_NULL, 1);
+    ProfileDatabase* db = (ProfileDatabase*)callback_params->db;
+    rocprofvis_db_ext_data_t record;
+    if (callback_params->future->Interrupted()) return SQLITE_ABORT;
+    record.category = "ARGS";
+    record.type = callback_params->future->GetProcessedRowsCount();
```
tomk-amd: Add comment that record.type does not indicate the type for "Argument" entries?

Or maybe create a new Adder function ex: FuncAddExtDataArgument?

`typedef rocprofvis_dm_result_t (*rocprofvis_dm_add_extdata_args_func_t) (const rocprofvis_dm_extdata_t object, rocprofvis_db_ext_data_t& name, rocprofvis_db_ext_data_t& value, rocprofvis_db_ext_data_t& type, uint8_t& position);`

vstempen: Agree, that's a bit confusing. I'll create FuncAddExtDataArgument

vstempen: Added new method

vstempen:
I didn't cleanup Callstack class in this PR. But you are right, we can probably clean it up. The reference to arguments was created to support RPD, but we don't have any RPD sample with callstack. I took it from a document Michael Wooton created a while ago. But I guess it never come to life. Was replaced with new schema instead.

tomk-amd (approved):

PR #590, opened 2026-01-05
Callstack Cleanup

drchen-amd:
-Display call stack information in table form instead of showing raw json blob.
-For events without call stack information, show `No Data Available` instead of a row of `{}`.
-Remove unused call stack property enums.

### DIFF
--- a/src/controller/src/rocprofvis_controller_event.cpp
+++ b/src/controller/src/rocprofvis_controller_event.cpp
@@ -275,7 +275,7 @@ Event::FetchDataModelStackTraceProperty(uint64_t event_id, Array& array,
                                             {
                                                 result = array.SetObject(kRPVControllerArrayEntryIndexed, entry_counter++, (rocprofvis_handle_t*) call_stack);
                                             }
-                                            else
+                                            if(result != kRocProfVisResultSuccess)
                                             {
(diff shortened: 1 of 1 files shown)


src/controller/src/rocprofvis_controller_event.cpp
```diff
@@ -247,14 +247,38 @@ Event::FetchDataModelStackTraceProperty(uint64_t event_id, Array& array,
...
+                                        if(!file.empty() || !pc.empty() || !name.empty() || !line_name.empty() || !line_address.empty())
+                                        {
+                                            CallStack* call_stack = new CallStack(file.c_str(), pc.c_str(), name.c_str(), line_name.c_str(), line_address.c_str());
+                                            result = array.SetUInt64(kRPVControllerArrayNumEntries, 0, entry_counter + 1);
+                                            if(result == kRocProfVisResultSuccess)
+                                            {
+                                                result = array.SetObject(kRPVControllerArrayEntryIndexed, entry_counter++, (rocprofvis_handle_t*) call_stack);
+                                            }
+                                            else
+                                            {
```
tomk-amd: Change `else` to `if (result != kSuccess)` ? the result in the inner block is not currently checked?

PR #593, opened 2026-01-06
Amokiche/Improve sidebar name labels

amokiche-amd:
## Motivation
* Sidebar
	- Improve name labels here as well, for example instead of:
		- Processes(1)
		 - [77777]
		  - Threads
		   - Thread: ./main.py(7777): Thread 7111 (1)
		
		it should be:
		
		- Processes(1)
		 - ./main.py (77777)
		  - Threads
		   - Thread 7111 (1)

## Technical Details
Change the logic of track name creation
image

## Submission Checklist

### DIFF
--- a/src/view/src/rocprofvis_track_item.cpp
+++ b/src/view/src/rocprofvis_track_item.cpp
@@ -40,6 +40,7 @@ TrackItem::TrackItem(DataProvider& dp, uint64_t id,
 , m_chunk_duration_ns(TimeConstants::ns_per_s * 30)  // Default chunk duration
 , m_tpt(tpt)  
 , m_track_project_settings(m_data_provider.GetTraceFilePath(), *this)
+, m_meta_area_label("")
 , m_pill("", false, false)
 {
     if(m_track_project_settings.Valid())
@@ -608,11 +609,10 @@ TrackItem::SetTrackName(const TrackInfo* track_info)
         {
             m_name += "Sampled ";
             std::string lower_case_sub_name = track_info->sub_name;
-            std::transform(
-                lower_case_sub_name.begin(), lower_case_sub_name.end(), lower_case_sub_name.begin(),
-                [](unsigned char c) {
-                return static_cast<char>(std::tolower(c));
-            });
+            std::transform(lower_case_sub_name.begin(), lower_case_sub_name.end(),
+                           lower_case_sub_name.begin(), [](unsigned char c) {
+                               return static_cast<char>(std::tolower(c));
+                           });
             m_name += lower_case_sub_name;
             break;
         }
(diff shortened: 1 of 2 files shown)


drchen-amd (commented):
Not sure if this matters but since this name is used in the meta area as well, but unlike the side bar, meta area has no nesting to tell what process a track is from.

So maybe if you had multiple processes, meta area becomes harder to tell threads from different processes apart?

Old vs New

Image

tomk-amd:
> Not sure if this matters but since this name is used in the meta area as well, but unlike the side bar, meta area has no nesting to tell what process a track is from.
> 
> So maybe if you had multiple processes, meta area becomes harder to tell threads from different processes apart?

This was my suggestion, to remove the process name from the meta area desc too.   One thing that I forgot to mention though was that the process name **should** remain on the thread that is marked as **MAIN**.

Perfetto for example does not show the process name for threads:

image

I feel spamming the process name for every thread track is too much? But happy to hear other opinions.

Maybe showing "Thread 154714 : 156619 (2)" would be a compromise  Thread proc_id : thread_id so that it still obvious which process the thread belonds to?

drchen-amd:
I think current look for this PR is fine enough. Maybe in the future we could put a colored dot on the sidebar at each process, and put that same colored dot in the meta area for all of a process' tracks.

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -262,8 +261,24 @@ TrackItem::RenderMetaArea()
...
+        if (m_show_pill_label)
+        {
+            const TrackInfo* track_info =
+                m_data_provider.DataModel().GetTimeline().GetTrack(m_id);
+
+            std::string main_track_name = m_data_provider.DataModel()
+                                    .GetTopology()
+                                    .GetProcess(track_info->topology.process_id)
+                                              ->command +
+                                          ": " + m_name;
```
drchen-amd: I think it's better if this track name string is stored, instead of building it every render. The m_show_pill_label is only set once.

src/controller/src/rocprofvis_controller_trace.h
```diff
@@ -40,7 +40,6 @@ class Trace : public Handle
                                  rocprofvis_db_progress_percent_t progress_percent,
                                  rocprofvis_db_status_t           status,
                                  rocprofvis_db_status_message_t message, void* user_data);
-
```
drchen-amd: Better to revert this and keep the file's history clean.

src/model/src/database/rocprofvis_db_sqlite.cpp
```diff
@@ -676,7 +676,7 @@ rocprofvis_dm_result_t  SqliteDatabase::ExecuteSQLQuery(DbInstance* db_instance,
     std::string guid_str = GuidAt(db_instance->GuidIndex());
     ReplaceAllSubstrings(query_str, "%GUID%", guid_str);
     query = query_str.c_str();
-    int   rc = Sqlite3Exec(conn, query, params->callback, params);
```
drchen-amd: Better to revert this and keep the file's history clean.

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -471,6 +464,171 @@ TrackItem::FetchHelper()
...
+                return static_cast<char>(std::tolower(c));
+            });
+            m_name += lower_case_sub_name;
+            break;
+        }
+        case TrackInfo::Topology::Counter:
+        {
+
+            m_name += track_info->main_name + ":" + track_info->sub_name;
+            break;
```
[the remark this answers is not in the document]
amokiche-amd: yes, but we can get the situation with different counters, but with the same name

PR #604, opened 2026-01-09
Use map instead of vector for CLI parser

tomk-amd:
Use map instead of vector for CLI parser.
Additional error checking added.


Sajeeth-Wimalasuriyan (approved):
Changes are good. Map should be faster too thanks tom.

PR #608, opened 2026-01-12
AIPROFVIS-25

Sajeeth-Wimalasuriyan:
Switched from showing the user eventid to database id. 

Ticket --> https://amd-hub.atlassian.net/browse/AIPROFVIS-25


src/view/src/rocprofvis_events_view.cpp
```diff
@@ -412,10 +407,15 @@ EventsView::HandleEventSelectionChanged(const uint64_t event_id, const bool sele
...
-                           "Event ID: " + std::to_string(event_data->basic_info.m_id),
-                           std::move(container), event_data, 0.0f });
+            // Get DB from event ID
+            TraceEventId tid;
+            tid.id         = event_data->basic_info.m_id;
+            uint64_t db_id = tid.bitfield.db_event_id;
+
+            m_event_items.emplace_front(EventItem{
+                m_event_item_id++, event_data->basic_info.m_id,
+                "ID: " + std::to_string(db_id), std::move(container), event_data, 0.0f });
```
tomk-amd: Don't add "ID: ", just put the number please

src/view/src/model/rocprofvis_model_types.h
```diff
@@ -78,9 +78,24 @@ struct EventExtData
...
+    struct
+    {
+        // Event ID from database
+        uint64_t db_event_id : 52;
+        // Node index
+        uint64_t event_node : 8;
+        // Event operation type
+        uint64_t event_op : 4;
+    } bitfield;
+    uint64_t id;
```
tomk-amd: There is a number of places that have statements like event.id.id which looks silly.

Let's change id -> uuid in the union, and rename db_event_id too make things consistent:

```suggestion
        // Event ID from database
        uint64_t event_id : 52;
        // Node index
        uint64_t event_node : 8;
        // Event operation type
        uint64_t event_op : 4;
    } bitfield;
    uint64_t uuid;
```

PR #620, opened 2026-01-15
Show Event Id from database instead of "Full" Event Id in tables

tomk-amd:
## Motivation

Show the database event id instead of the application event id (decorated id) in table views.  

## Technical Details

In the model a new column "DB Id" was added to table queries.  Existing Id column's public name renamed to "__id".  New column now uses "id" as public name.

UI updated to recognize new names.

### DIFF
--- a/src/view/src/rocprofvis_event_search.cpp
+++ b/src/view/src/rocprofvis_event_search.cpp
@@ -17,7 +17,7 @@ namespace View
 constexpr uint64_t    MAX_RESULTS_DISPLAYED = 5;
 constexpr const char* TRACK_ID_COLUMN_NAME  = "__trackId";
 constexpr const char* STREAM_ID_COLUMN_NAME = "__streamTrackId";
-constexpr const char* ID_COLUMN_NAME        = "__id";
+constexpr const char* ID_COLUMN_NAME        = "__uuid";
 constexpr const char* EVENT_ID_COLUMN_NAME  = "id";
 constexpr const char* NAME_COLUMN_NAME      = "name";
 
--- a/src/model/src/database/rocprofvis_db_query_builder.h
+++ b/src/model/src/database/rocprofvis_db_query_builder.h
@@ -164,7 +164,7 @@ class Builder
         static constexpr const char* END_SERVICE_NAME       = "endTs";
         static constexpr const char* START_PUBLIC_NAME     = "start";
         static constexpr const char* END_PUBLIC_NAME       = "end";
-        static constexpr const char* ID_PUBLIC_NAME = "__id";
+        static constexpr const char* ID_PUBLIC_NAME = "__uuid";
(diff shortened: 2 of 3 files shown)


src/view/src/widgets/rocprofvis_infinite_scroll_table.h
```diff
@@ -36,6 +36,7 @@ class InfiniteScrollTable : public RocWidget
     enum ImportantColumns
     {
         kId,
+        kEventId,
```
drchen-amd: I think it would be better if they are called db id and uuid?

src/model/src/database/rocprofvis_db_table_processor.cpp
```diff
@@ -902,6 +902,10 @@ namespace DataModel
                 value.bitfield.event_node = callback_params->db_instance->GuidIndex();
                 value.bitfield.event_op = op;
                 table_processor->m_tables[callback_params->track_id]->PlaceValue(column_index, value.value);
+            } else if (columns[column_index].m_schema_index == Builder::SCHEMA_INDEX_EVENT_DB_ID)
+            {
+                uint64_t id = db->Sqlite3ColumnInt64(func, stmt, azColName, columns[column_index].m_orig_index);
+                table_processor->m_tables[callback_params->track_id]->PlaceValue(column_index, id);
             }
```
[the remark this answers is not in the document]
vstempen: Right. We need to have unconditional else there. And, as we discussed with Tom offline, may be better to unconditionally put "table_processor->m_tables[callback_params->track_id]->PlaceValue(DB_ID_PUBLIC_NAME, id);" into the "if (columns[column_index].m_schema_index == Builder::SCHEMA_INDEX_EVENT_ID)" block instead of reading id twice from SQL.

src/view/src/widgets/rocprofvis_infinite_scroll_table.h
tomk-amd: Updated the names, chose DbEventId for the pure database one

vstempen (approved):

src/view/src/rocprofvis_event_search.cpp
```diff
@@ -17,7 +17,8 @@ namespace View
 constexpr uint64_t    MAX_RESULTS_DISPLAYED = 5;
 constexpr const char* TRACK_ID_COLUMN_NAME  = "__trackId";
 constexpr const char* STREAM_ID_COLUMN_NAME = "__streamTrackId";
-constexpr const char* ID_COLUMN_NAME        = "id";
+constexpr const char* ID_COLUMN_NAME        = "__id";
+constexpr const char* EVENT_ID_COLUMN_NAME  = "id";
```
drchen-amd: These should get the uuid and db id naming as well.

PR #625, opened 2026-01-19
MiniMap Counter Support

Sajeeth-Wimalasuriyan:


vstempen:
There is a value and density in histogram table now. It seems you are using value for density, which is not right. You should use kRPVControllerTrackHistogramBucketDensityIndexed to get density values and kRPVControllerTrackHistogramBucketValueIndexed for value. You should use density for events and counters colors, but value for counter mini-map bar height.

PR #626, opened 2026-01-19
Compute View Skeleton

drchen-amd:
Initial commit for compute trace view to facilitate testing interface and serve as foundation for further development.
- Moved compute related sources into `src/compute/`
- Moved system trace specific options in `Edit` menu into `TraceView::RenderEditMenuOptions()`
- Moved loading screen rendering into `TraceView` into `RootView::RenderLoadingScreen()`.
- Setup `ComputeView` skeleton derived from `RootView` following `TraceView` patterns.
- Setup skeleton `DataProvider` path for loading compute traces.

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -575,8 +575,7 @@ AppWindow::RenderViewMenu(Project* project)
                         kRocProfVisViewNotification_Toggle_Fullscreen);
             }
         }
-        ImGui::Separator();
-        ImGui::MenuItem("ROCm(TM) Systems Profiler", NULL, false, false);
+        ImGui::SeparatorText("System Profiler Panels");
 #endif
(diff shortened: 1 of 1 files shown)


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -586,6 +565,18 @@ AppWindow::RenderViewMenu(Project* project)
                 tool_bar_item->m_visible = settings.show_toolbar;
             }
         }
+#ifdef COMPUTE_UI_SUPPORT
+        if(ImGui::MenuItem("Fullscreen", "F11", m_is_fullscreen))
+        {
+            if(m_notification_callback)
```
tomk-amd: The fullscsreen menu item will appear in a different place when compute support is enabled / disabled.  Not a big deal though, as once compute is ready all of these #ifdef statements will be removed and this will have to be cleaned up anyway.

PR #651, opened 2026-01-28
Sample table grouping enabled, group filtering removed, other UI clean up

tomk-amd:
## Motivation

Rename some of the labels and wording in the UI.
Clean up table filter interface.

## Technical Details

The updated model behaviour no longer supports (intuatively) the "where filter" when grouping is enabled, so now this is disabled  when group by column is selected.
The group by column filter box has been completely removed (except in developer mode).

The group by option has been enabled for Sample tables.

Mini map labels and Timeline context menu options have been renamed / reworded.

### DIFF
--- a/src/model/src/database/rocprofvis_db_profile.cpp
+++ b/src/model/src/database/rocprofvis_db_profile.cpp
@@ -980,17 +980,19 @@ ProfileDatabase::BuildTableQuery(
     table_string_id_filter_map_t string_id_filter_map;
     std::string group_by_select;
     std::string group_by;
-    if(summary)
+    
+    bool sample_query = false;
+    if(TABLE_QUERY_UNPACK_OP_TYPE(tracks[0]) == 0)
     {
-        bool sample_query = false;
-        if(TABLE_QUERY_UNPACK_OP_TYPE(tracks[0]) == 0)
-        {
-            sample_query = TrackPropertiesAt(tracks[0])->process.category == kRocProfVisDmPmcTrack;
-        }
-        else
-        {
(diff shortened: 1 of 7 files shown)


vstempen (approved):
Data model change is good.

drchen-amd (changes requested):
This needs handling for the scenario where a filter is entered, then grouping is enabled after. Currently this will result in populated but disabled filter box that still takes effect in the query.
image

drchen-amd:
> This needs handling for the scenario where a filter is entered, then grouping is enabled after. Currently this will result in populated but disabled filter box that still takes effect in the query. image

#653

tomk-amd:
> This needs handling for the scenario where a filter is entered, then grouping is enabled after. Currently this will result in populated but disabled filter box that still takes effect in the query. image

Hmm - I made a change to clear the fliter when grouping.. but after playing with it a bit more, maybe there is value in keeping the filter.. for example if the user sets a filter like "duration > 10000" and then runs grouping they will get aggregations that run only on values that meet the requirement.  Maybe there are situations where this is actually useful?  Not letting them change the filter while in group mode I think makes sense as they no longer have a visual reference to the orginal columns, but maybe allowing it to stay applied isn't so bad?

drchen-amd (approved):

tomk-amd:
> Its up to you. I did not note down specifically what I grouped by and my filter, but I got a couple of crashes when trying it out as is. Forcing only grouping or filtering would lessen the likelihood of some combination being entered that causes a crash.

Yeah after some more testing I agree.  Disabled the filter when grouping is active.

Sajeeth-Wimalasuriyan (approved):
Changes look good. I don't see any issues. I also ran and verified everything on my system — good to go.

PR #654, opened 2026-01-30
Annotation Bug Fix

Sajeeth-Wimalasuriyan:


src/view/src/rocprofvis_trace_view.cpp
```diff
@@ -691,10 +691,13 @@ TraceView::RenderAnnotationControls()
     ImGui::PushID("add_new_sticky");
     if(ImGui::Button(ICON_ADD_NOTE))
     {
-        ViewCoords coords = m_timeline_view->GetViewCoords();
+        ViewCoords coords       = m_timeline_view->GetViewCoords();
+        ImVec2     graph_size   = m_timeline_view->GetGraphSize();
+        double     center_time  = m_timeline_view->GetTransform()->PixelToTime(graph_size.x * 0.5f);
```
tomk-amd: Paranoid - should check pointer is valid before using.  It'll work now, but if timeline ever changes its implemenation it could return a null pointer.

PR #661, opened 2026-02-06
Fixed track id tracking

vstempen:
## Motivation

Incorrect table content when unselecting one of the tracks with shared events.
Event rows may have wrong track IDs

## Technical Details

Reworked track ID tracking so it can be run per each row. Before track ID was detected once per request, but it does not work for stream tracks or for dispatch tracks which have events from different streams.

Created a class for fast lookup of track ID based on chosen track identifiers
Removed multiple code handling ID tracking, replaced all ID tracking by single class.


tomk-amd:
Seems to work correctly now

tomk-amd (approved):

PR #668, opened 2026-02-11
Update optiq-install.rst

mattwill-amd:
Update download links


tomk-amd:
Merged manually instead

PR #670, opened 2026-02-12
Add Summary view for compute

amokiche-amd:
## Motivation
Add Summary view for compute view

image

## Technical Details
Added summary view with kernel table and two charts
small refactoring compute tester

## Submission Checklist


drchen-amd (commented):
I think something is wrong with the metric dropdown for the charts. The chart looks the same for all the duration, but looking at the table the total duration should be different:

Image

Image

drchen-amd:
I think the charts should be styled like system trace summary view charts. It would be nice if all of our charts throughout the app supported a common baseline feature set (hover, tooltip...etc) and style for consistency.  You would also get the long name handling for free.

tomk-amd:
Reworked via different PR

PR #680, opened 2026-02-19
Added Value Names List to Enable Query Builder

Sajeeth-Wimalasuriyan:


src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -149,6 +150,22 @@ namespace DataModel
...
+			query += "cmv.metric_id as sub_table_id, ";
+			query += "cmv.value_name ";
+			query += "FROM compute_metric_view cmv ";
+			query += "INNER JOIN compute_metric_definition cmd ON cmv.metric_id = cmd.metric_id ";
+			query += "WHERE cmd.workload_id = ";
+			query += params[0].param_str;
+		}
+		return query;
+	}
+
```
drchen-amd: This is your query:
Image

In this example there are 6471 results, and each result is stored into the workload. I'm not a fan of this approach because the vast majority of these results are duplicates. There are only 24 unique value names for this workload. The task of removing the duplicates falls downstream into your DataProvider, where you search and discard duplicates.
 
I suggest that the query returns only unique values, such that the workload does not waste space storing duplicates, and the UI does not need to worry about detecting and discarding duplicates.

tomk-amd: I was just looking at this too.

I think the controller should run the query for each sub-table type and  the query should be like the one Derek provided in the chat:

```
SELECT DISTINCT value_name
FROM compute_metric_view
WHERE metric_id LIKE '**2.1.%**' and workload_id = **1**
```
The **bolded** items are parameters to the query.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -3874,6 +3874,40 @@ DataProvider::ProcessLoadComputeTrace(RequestInfo& req)
...
+            auto cat_it =
+                workload.available_metrics.tree.find(static_cast<uint32_t>(cat_id));
+            if(cat_it != workload.available_metrics.tree.end())
+            {
+                auto tbl_it =
+                    cat_it->second.tables.find(static_cast<uint32_t>(tbl_id));
+                if(tbl_it != cat_it->second.tables.end())
+                {
+                    auto& vn = tbl_it->second.value_names;
+                    if(std::find(vn.begin(), vn.end(), name) == vn.end())
```
drchen-amd: Confirm if still needed.

src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -149,6 +150,22 @@ namespace DataModel
...
+		std::string query;
+		if (num == 1 && params != nullptr && params[0].param_type == kRPVComputeParamWorkloadId)
+		{
+			query = "SELECT DISTINCT ";
+			query += "substr(cmv.metric_id, 0, instr(cmv.metric_id, '.')) as table_id, ";
+			query += "cmv.metric_id as sub_table_id, ";
+			query += "cmv.value_name ";
+			query += "FROM compute_metric_view cmv ";
+			query += "INNER JOIN compute_metric_definition cmd ON cmv.metric_id = cmd.metric_id ";
+			query += "WHERE cmd.workload_id = ";
```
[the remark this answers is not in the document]
drchen-amd: ^

drchen-amd (changes requested):

PR #682, opened 2026-02-20
Centralized workload / kernel selection tracking

tomk-amd:
## Motivation

This PR introduces a centralized selection management system for the compute UI.
Emits selection change events.
Emits metrics fetched event.

## Technical Details

Refactored ComputeKernelDetailsView and ComputeMemoryChartView to use new ComputeSelection class and Events.

Metric fetching and data updates are now event-driven.


drchen-amd (approved):
This isn't from this change but there are some overlap issues in the memory chart depending on font size:

Image

PR #693, opened 2026-02-25
Table Widget 

Sajeeth-Wimalasuriyan:


src/view/src/compute/rocprofvis_compute_kernel_details.cpp
```diff
@@ -116,12 +105,49 @@ ComputeKernelDetailsView::Render()
...
+
+    uint32_t wid = m_compute_selection->GetSelectedWorkload();
+    uint32_t kid = m_compute_selection->GetSelectedKernel();
+    if(wid == ComputeSelection::INVALID_SELECTION_ID ||
+       kid == ComputeSelection::INVALID_SELECTION_ID)
+        return;
+
+    m_data_provider.FetchMetrics(MetricsRequestParams(
+        wid, {kid}, {{2, 1, std::nullopt}}, m_client_id));
+}
```
tomk-amd: Magic numbers...

src/view/src/compute/rocprofvis_compute_kernel_details.h
```diff
@@ -26,11 +27,15 @@ class ComputeKernelDetailsView : public RocWidget
     void Update() override;
 
 private:
+    void FetchSOLMetrics();
+    void UpdateSOLTable();
+
```
tomk-amd: Could this functionality be moved into the Widget?  It would need a pointer or reference to the Compute Model,  but then client could call `m_sol_table.FetchMetrics(workload_id, kernel_id, cat_id, table_id)` and `m_sol_table.UpdateTable()`

tomk-amd (approved):
This is fine for now.  It will need to be reworked once the kernel selection PR is merged.  And once fetching metrics in parallel is implemented.

PR #695, opened 2026-02-25
Fixed Warnings Using Cursor 

Sajeeth-Wimalasuriyan:


src/view/src/rocprofvis_minimap.cpp
```diff
@@ -122,6 +121,7 @@ Minimap::SetData(const std::vector<std::vector<double>>& data)
     }
 
     const size_t height = data.size();
+    (void)height;
```
tomk-amd: if height is not used just remove it -it is not a parameter to the function, it is defined right here

PR #696, opened 2026-02-26
Flexbox for OPTIQ

Sajeeth-Wimalasuriyan:


src/view/src/widgets/rocprofvis_flex_container.h
```diff
@@ -0,0 +1,32 @@
...
+    float                      height    = 0.0f;
+    float                      flex_grow = 1.0f;
+};
+
+class FlexContainer : public RocWidget
+{
+public:
+    void Render() override;
+
+    std::vector<FlexItem> items;
```
tomk-amd: Would be nice to have some getter and setter functions for adjusting the properties of an item.  For example to maximize / shrink the roof line chart.

Ex:

m_flex->GetItem(roofline_id).min_width = 1000,0f;

tomk-amd (changes requested):

PR #697, opened 2026-02-26
Add kernel selection table

tomk-amd:
## Motivation

Add a kernel selection table.  
Table allows user to add / remove custom metric value columns.
The active kernel can be selected using this table.

Also added support for running metric fetches in parallel. (This fixes issue where the table view does not update correctly when switching kernels because fetch request gets blocked by another client).

## Technical Details

Metrics are selected using the query builder helper object.
Kernel selection is synchronized with toolbar kernel selection combo box.
Additional utility functions added to ComputeDataModel.
Compute tester moved to its own implementation files.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
+++ b/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
@@ -361,11 +361,6 @@ KernelMetricTable::Render()
 
                 ImGui::EndTable();
             }
-
-            if(request_pending)
-            {
-                RenderLoadingIndicator();
-            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+                    // Sortable header with X button
+                    const char* name = ImGui::TableGetColumnName(col);
+                    ImGui::TableHeader(name);
+                    ImVec2 text_size = ImGui::CalcTextSize(name);
+                    ImGui::SameLine(text_size.x, item_spacing);
+
+                    ImGui::PushID(col);
+                    ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(0.0f, 0.0f));
+                    if(ImGui::SmallButton("X"))
```
drchen-amd: This should be ICON_X_CIRCLED like the system trace selection removals.

src/view/src/widgets/rocprofvis_query_builder.cpp
drchen-amd: Can be exceeded and cut off the buttons at large font sizes.
Image

PR #698, opened 2026-02-27
Compute Top Kernels

drchen-amd:
`TimelineView`
- Factored out grid/histogram ruler interval calculation into `FittedGraphAxisInterval()` in `rocprofvis_utils.cpp`. The top kernels bar chart makes use of it.

`ComputeDataModel`
- Moved Kernel invocation count and duration metrics into array for more convenient access.

`ComputeSummaryView`

- Added `ComputeTopKernels` widget

`Roofline`
- Reworked the conditions for "No Data Available" screen. Widget will always render its frame regardless of `ComputeSelection`. This was a small change but resulted in a large diff in `rocprofvis_compute_roofline.cpp` as `{}` where moved around.
- Reworked the logic for differentiating between "single kernel" and "all kernels" modes. Instead of being implicit based on SetKernel(), it is now explicitly set on creation. This removes the previous behavior of re-build the workload metadata when only the kernel selection changes.
- Corrected typo `Inovcation(s)` in kernel tooltip.
- Added missing unit for `Duration `in kernel tooltip.

`SummaryView`
- Fixed bar chart hover region being wider than bar width.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_roofline.cpp
+++ b/src/view/src/compute/rocprofvis_compute_roofline.cpp
@@ -754,7 +754,7 @@ Roofline::RenderMenus(const ImVec2 region, const ImGuiStyle& style,
             ImGui::Checkbox("##scale_intensity", &m_scale_intensity);
             ImGui::SameLine();
             ElidedText("Scale kernel marker size to duration",
-                       ImGui::GetContentRegionAvail().x, region.x * 0.5f, true);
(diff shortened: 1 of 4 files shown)


drchen-amd:
> Can table columns be sorted? Clicking on the headers does not initiate a sort?

I was not planning to have any sorting besides the static top 10 by duration since any sorting would be done temporarily by the view. I can do it anyway if you believe it worth it.

PR #701, opened 2026-02-28
Table Enhancement v2

Sajeeth-Wimalasuriyan:
Added metric ID column, description tooltip on hover, and right-click copy support to the compute table view


tomk-amd:
The description tool tip needs to have a max width limit, and allow multi-line text wrapping: 
image

See meta area tooltip in `TrackItem::RenderMetaArea()` as an example.

PR #710, opened 2026-03-04
Add UI for filtering kernel selection table

tomk-amd:
## Motivation

Add UI for filtering kernel selection table

## Technical Details

- query in model updated
- controller interface updated to handle filter query params
- ui updated to add row of text inputs for filter expressions
- enable horizontal scolling on kernel selection table
- update table height calculation


drchen-amd (changes requested):
I don't think the column widths should scale unbounded because some kernel names are very long. 

There is also what I suspect is some sort of auto sizing feedback loop going on when you switch workloads and the column sizes slowly update. If I had to guess it may be ColumnFilter and the column itself both auto sizing to each other. 

[Screencast from 03-04-2026 01:18:28 PM.webm](https://github.com/user-attachments/assets/8d3d24eb-ad51-4e14-9c79-6bf2971c6aac)

drchen-amd (approved):

PR #739, opened 2026-03-20
Use Compute Table Interface in Systems 

Sajeeth-Wimalasuriyan:
image


drchen-amd (commented):
Filtering does not replace grouping so I don't think the grouping should be removed.

PR #746, opened 2026-03-25
Added Database Clean Feature 

Sajeeth-Wimalasuriyan:
image


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -649,6 +721,53 @@ AppWindow::RenderFileMenu(Project* project)
...
+
+            std::string project_id = has_trace ? project->GetID() : "";
+            auto start_cleanup = [this, trace_view, project_id](bool rebuild) {
+                m_disable_app_interaction = true;
+                m_cleanup_in_progress     = true;
+                trace_view->CleanupDatabase(rebuild, [this, project_id]() {
+                    m_cleanup_in_progress     = false;
+                    m_disable_app_interaction = false;
+                    m_tab_container->RemoveTab(project_id);
+                });
```
tomk-amd: I think clean up progress should be tracked on a per-trace (db) level.  It does not need to block the entire application?   This logic and the cleanup indicator splash screen should probably be moved to trace view.  We can do this in a separate PR.

tomk-amd (changes requested):

tomk-amd (approved):

PR #758, opened 2026-04-01
Populate kernel selection table from table view

amokiche-amd:
## Motivation
Improve the compute UI workflow and simplify navigation between related views. This branch makes metric exploration more direct and reduces friction when moving from summary-style tables into detailed kernel analysis.

## Technical Details
branch refactors parts of the compute view interaction model to use a cleaner event-driven flow between components. It also restructures metric table handling to share more logic, improve maintainability, and support richer table actions consistently across the compute UI.

In addition, the branch includes supporting cleanup in metric identification/data lookup and small UI polish updates, including improved menu/icon rendering.
image

## Submission Checklist


drchen-amd (changes requested):
Crashes when switching workload/kernel if there is a pinned metric whose category is not the active tab. For example:

1. Go to Speed of light tab and pin something.
2. Switch Workload/kernel
3. Switch to any other tab. (Crashes)

Image

src/view/src/compute/rocprofvis_compute_view.cpp
```diff
@@ -119,11 +118,10 @@ void
 ComputeView::CreateView()
 {
     m_compute_selection = std::make_shared<ComputeSelection>(m_data_provider);
-
     m_tab_container = std::make_shared<TabContainer>();
     m_tab_container->AddTab(TabItem{"Summary View", "compute_summary_view", std::make_shared<ComputeSummaryView>(m_data_provider, m_compute_selection), false});
-    m_tab_container->AddTab(TabItem{"Kernel Details", "compute_kernel_details_view", std::make_shared<ComputeKernelDetailsView>(m_data_provider, m_compute_selection), false});
-    m_tab_container->AddTab(TabItem{"Table View", "compute_table_view", std::make_shared<ComputeTableView>(m_data_provider, m_compute_selection), false});
```
drchen-amd, line 126: This file should be reverted.

src/view/src/compute/rocprofvis_compute_view.h
```diff
@@ -7,6 +7,8 @@
 #include "widgets/rocprofvis_query_builder.h"
 #include "widgets/rocprofvis_tab_container.h"
 #include "rocprofvis_compute_selection.h"
+#include "rocprofvis_compute_kernel_details.h"
+
```
drchen-amd: This file should be reverted.

src/view/src/rocprofvis_trace_view.cpp
```diff
@@ -711,7 +711,7 @@ TraceView::RenderToolbar()
     ImGui::PopStyleColor(4);
     if(ImGui::IsItemHovered())
     {
-        SetTooltipStyled("Reset view to default zoom and pan");
```
drchen-amd, line 714: This file should be reverted.

src/view/src/icons/rocprovfis_icon_defines.h
```diff
@@ -26,29 +26,37 @@ constexpr ImWchar icon_ranges[] = {
...
+inline constexpr char ICON_ADD_NOTE[]           = u8"\uF218";
+inline constexpr char ICON_ARROWS_SHRINK[]      = u8"\uF267";
+inline constexpr char ICON_COMPASS[]            = u8"\uF273";
+inline constexpr char ICON_CHART_BAR[]          = u8"\uF2B5";
+inline constexpr char ICON_DELETE[]             = u8"\uF2D7";
+inline constexpr char ICON_EYE_SLASH[]          = u8"\uF306";
+inline constexpr char ICON_TREE[]               = u8"\uF33F";
+inline constexpr char ICON_GRID[]               = u8"\uF35C";
+inline constexpr const char* ICON_ARROW_DOWN    = u8"\uF35D";
+inline constexpr const char* ICON_ARROW_UP      = u8"\uF366";
```
drchen-amd, line 29: These do not follow the pattern.

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,361 +3,440 @@
...
+    m_columns[1]          = "Metric ID";
+    m_columns[2]          = "Metric";
+    m_lust_column_index   = 3;
+    m_columns[LAST_INDEX] = "Unit";
+}
+
+bool
+MetricTableBase::IsMetricPined(MetricId metric_id)
+{
+    return m_rows[metric_id].pinned;
```
drchen-amd: This will insert if metric id is not in map.

drchen-amd (changes requested):
Will require a second look once crashing is resolved.

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,361 +3,440 @@
...
+        SectionTitle(m_table_title.c_str());
+
+    if(m_rows.empty())
     {
-        for(const auto& vn : table.value_names)
-            m_column_names.push_back(vn);
+        ImGui::TextDisabled("Pin the metric to see it here");
+        return;
     }
-    m_column_names.push_back("Unit");
```
tomk-amd, line 40: I don't think this message should be displayed by the base class.  It is specialized for Pinned tables...

src/view/src/model/compute/rocprofvis_compute_model_types.h
```diff
@@ -131,25 +131,60 @@ struct MetricValue
...
+    {
+        return std::to_string(category_id) + "." + std::to_string(table_id) + "." +
+               std::to_string(entry_id);
+    }
+
+    bool operator==(const MetricId& other) const noexcept
+    {
+        return category_id == other.category_id && table_id == other.table_id &&
+               entry_id == other.entry_id;
+    }
```
tomk-amd, line 149: I'm not sure the MetricId struct is preferrable over the simple the union MetricId (MetricKey)...  All of the additional functionality provided by the struct like the ToString() function could be added to the model as a static function for example: std::string ComputeDataModel::MetricIdToString(MetricKey key) and even MetricIdToString(uint64_t metricId). 

Also in the new event "ComputeAddMetricToKernelDetailsEvent" instead of passing in a struct for the id, a uint64_t could be used.  This would also get rid of the compute_types.h depemdcy creeping into the events.h file and by extension to everyone that includes it.

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,361 +3,440 @@
...
+            }
+            else
+            {
+                label = " " + std::string(ICON_CHEK_BOX_CHECKED) + " Unpin metric";
+                if(ImGui::MenuItem(label.c_str()))
+                {
+                    row.second.pinned = !row.second.pinned;
+                    m_pin_metric_clicked({ row.first.category_id, row.first.table_id,
+                                           row.first.entry_id });
+                }
```
tomk-amd: Not sure we need pinning via menu anymore, since the pin column has been added

src/view/src/icons/rocprovfis_icon_defines.h
amokiche-amd: What a patern do you mean?

src/view/src/model/compute/rocprofvis_compute_model_types.h
amokiche-amd: You're the boss and the final decision is yours, but I see the following problems with MetricKey:
The following properties of bit-fields are implementation-defined:

1. Depending on the implementation, the compiler may use different alignment policies, and the compilation result may differ depending on the compiler version and platform. 
cppreference:
The value that results from assigning or initializing a signed bit-field with a value out of range, or from incrementing a signed bit-field past its range.
Everything about the actual allocation details of bit-fields within the class object.
https://en.cppreference.com/w/cpp/language/bit_field.html?utm_source=chatgpt.com

2. C++ Core Guidlines recomend
C.183: Don’t use a union for type punning

3. in compute_data_model.cpp:158

Image
we initialize 16 bits fields by uint32_t parameters. 
Conclusion:
What do we gain by using a union? We save 4 bytes, but we deviate from the guidelines and introduce potential undefined behavior. MetricId uses 3 uint32_t fields and is only 4 bytes heavier, but it eliminates these risks. Considering all of the above, I would still try to get rid of the union.

src/view/src/rocprofvis_trace_view.cpp
amokiche-amd: unexpected changes :) thanks

src/view/src/icons/rocprovfis_icon_defines.h
amokiche-amd: Replacing it with an array just makes it more obvious, but I can agree that it might be better to go back to const char *, but adding inline saves us from unnecessary duplicates in every translation unit that includes this header.

src/view/src/model/compute/rocprofvis_compute_model_types.h
tomk-amd: Yeah I was thinking about this a bit more .. 

What we gain is some performance as a uint can be used as a map key instead of calling a custom hashing function. (although in our case this is probably negligable).

I am not concerned about point 1 as this item will never be shared across modules, especially ones built by different compilers.   

I agree with point 3, I am not entirely happy about converting 32 bit values to 16 bit ones.  Even though we do not expect these values to ever overflow in a 16 bit space it is still messy.  (Although if we really think that these values may overflow the 16bit boundary the hashing function will have the same problem and potentially have collisions).

One other concern was "spilling" compute dependencies into the rest of the application as mentioned in event example.. But this can be addressed by making the event take 3 params, category, table, entry ids instead of the struct?

src/view/src/icons/rocprovfis_icon_defines.h
tomk-amd: yes I think let's go back to char *, this is the "pattern" used throughout the application.  Inline is fine.

tomk-amd (approved):

src/view/src/widgets/rocprofvis_compute_widget.cpp
tomk-amd: I've approved the PR, but please change this to:

```
    auto it = m_rows.find(metric_id);
    return it != m_rows.end() && it->second.pinned;
```

drchen-amd:
> Will require a second look once crashing is resolved.

This crashing seems fixed.

drchen-amd (changes requested):
Seeing two other issues:

1. Metrics appear in random order because the rendering is done based on map order.
2. Crash during workload switch if a pinned metric is not present in the new workload.

Demo of both:

[Screencast from 04-16-2026 09:00:13 PM.webm](https://github.com/user-attachments/assets/e8e7fe01-cc09-410a-8474-59ae96075d56)

(For 2, I'm pinning 2.1.33, which is only in one of the workloads)

amokiche-amd:
> 

Thanks, great find, my mistake, now fixed.

drchen-amd:
> Seeing two other issues:
> 
>     1. Metrics appear in random order because the rendering is done based on map order.
> 
>     2. Crash during workload switch if a pinned metric is not present in the new workload.
> 
> 
> Demo of both:
> Screencast.from.04-16-2026.09.00.13.PM.webm
> 
> (For 2, I'm pinning 2.1.33, which is only in one of the workloads)

Fixed

tomk-amd:
The "Pin" column shows up on tables where pinning should not be enabled ex: Kernel SOL Table or Workload SOL Table

image


image

tomk-amd (changes requested):

drchen-amd (changes requested):
Context menu on linux has a d symbol:
Image

I confirmed on two machines using [workflow](https://github.com/ROCm/roc-optiq/actions/runs/24578440535) to ensure its not just my environment. Problem is isolated to this particular icon.

src/view/src/compute/rocprofvis_compute_view.cpp
drchen-amd: Not resolved.

drchen-amd:
> Context menu on linux has a d symbol: Image
> 
> I confirmed on two machines using [workflow](https://github.com/ROCm/roc-optiq/actions/runs/24578440535) to ensure its not just my environment. Problem is isolated to this particular icon.

Fixed

drchen-amd (changes requested):
N/A metrics are not handled correctly:

1. They cannot be pinned.
2. Their values become stale when switching from a kernel/workload that does not have N/A for that metric.
3. Their columns get merge with other pinned metrics that  are not N/A.

https://github.com/user-attachments/assets/5f168d52-0b8c-4ac9-8c20-ef360586bd8d

Also there is now an extra empty column in Kernel Details > System Speed of Light

tomk-amd:
> N/A metrics are not handled correctly:
> 
> 1. They cannot be pinned.

N/A columns should be pinnable, but this has to be done carefully.  For the "normal" metric tables pinning a column that has N/A as a value is ok. It is by defintion defined for that metric. It simply means that these values were not collected for the selected kernel.

However, in the pinned table (top table) some of the columns will always be N/A for certain metrics and they should not be pinnable.  These are columns that are the result of a "join" of metrics that come from tables that contain different columns.

Ex:  The metric 2.1.1 definition does not have Min and Max values and 12.2.4 does not have Pct of Peak and Peak values.

image

These columns should not be allowed to be sent to kernel details table.

> 2. Their values become stale when switching from a kernel/workload that does not have N/A for that metric.
> 3. Their columns get merge with other pinned metrics that  are not N/A.

Regarding 3, It looks like the columns are correctly added (merged) but the newly selected metric does not get added because all of its values are N/A.  

Ex:  I've selected 2 metrics but they do not show up in the pinned table (they should).

image

> Also there is now an extra empty column in Kernel Details > System Speed of Light

This is also true on the Summary page speed of light table.  It looks like the "Pin" column might still be counted when the table is being setup?

tomk-amd:
> This is also true on the Summary page speed of light table. It looks like the "Pin" column might still be counted when the table is being setup?

As suspected, the code here needs to account for the missing pin column https://github.com/ROCm/roc-optiq/pull/758/changes#diff-215f528b1050a0d87d910d04a3e772ad0d2b41a49e34b67fc2555513a6c9c78eR51

This fixes it:

```
    int num_columns = static_cast<int>(m_columns.size());
    if(!CanBePinned())
    {
        if(num_columns > 0) 
             num_columns--;
        if(num_columns == 0)
        {
            RenderEmptyTable();
            return;
        }
    }

    if(ImGui::BeginChild(("##" + m_table_title + "layout").c_str(),
                         ImVec2(0, GetTableHight()), false,
                         ImGuiWindowFlags_HorizontalScrollbar))
    {
        if(ImGui::BeginTable(("##" + m_table_title).c_str(), num_columns,
                              m_table_flags))
        {

```

amokiche-amd:
> N/A metrics are not handled correctly:
> 
> 1. They cannot be pinned.
> 2. Their values become stale when switching from a kernel/workload that does not have N/A for that metric.
> 3. Their columns get merge with other pinned metrics that  are not N/A.
> 
>  Screencast.from.04-20-2026.04.53.24.PM.webm 
> Also there is now an extra empty column in Kernel Details > System Speed of Light

fixed

amokiche-amd:
> > N/A metrics are not handled correctly:
> > 
> > 1. They cannot be pinned.
> 
> N/A columns should be pinnable, but this has to be done carefully. For the "normal" metric tables pinning a column that has N/A as a value is ok. It is by defintion defined for that metric. It simply means that these values were not collected for the selected kernel.
> 
> However, in the pinned table (top table) some of the columns will always be N/A for certain metrics and they should not be pinnable. These are columns that are the result of a "join" of metrics that come from tables that contain different columns.
> 
> Ex: The metric 2.1.1 definition does not have Min and Max values and 12.2.4 does not have Pct of Peak and Peak values.
> 
> image
> These columns should not be allowed to be sent to kernel details table.
> 
> > 2. Their values become stale when switching from a kernel/workload that does not have N/A for that metric.
> > 3. Their columns get merge with other pinned metrics that  are not N/A.
> 
> Regarding 3, It looks like the columns are correctly added (merged) but the newly selected metric does not get added because all of its values are N/A.
> 
> Ex: I've selected 2 metrics but they do not show up in the pinned table (they should).
> 
> image
> > Also there is now an extra empty column in Kernel Details > System Speed of Light
> 
> This is also true on the Summary page speed of light table. It looks like the "Pin" column might still be counted when the table is being setup?

image


image

now user can send to the kernel details metric if it empty for it kernel, and not a product of metric merging

drchen-amd (changes requested):
Removing pinned metrics breaks the column management:

https://github.com/user-attachments/assets/e191b036-b057-426e-bc0c-1cd9683432fa

amokiche-amd:
> Removing pinned metrics breaks the column management:
> 
>  Screencast.from.04-21-2026.01.15.12.PM.webm

Fixed, redisign the logic to add and remove new metric and removed potential race condition

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -433,12 +434,44 @@ PinnedMetricTable::PinnedMetricTable(DataProvider&                     data_prov
...
+
+        UpdateColumns(metric_id, new_columns, new_last_column_index);
+        FillTableRow(metric_id, new_columns, new_rows);
+    }
+
+    m_rows.swap(new_rows);
+    m_columns.swap(new_columns);
+    m_last_column_index = new_last_column_index;
+    m_rebuild_pending = false;
+}
```
tomk-amd, line 474: Maybe clear m_pending_pinned_ids here too?

tomk-amd (approved):

tomk-amd (commented):
The font merging code needs to be reverted..  Other previously working icons are being corrupted on linux (ubuntu).

tomk-amd (changes requested):

drchen-amd (changes requested):
Icon issue is still present, system trace annotation controls has a 3 symbol:

Image

amokiche-amd:
> Icon issue is still present, system trace annotation controls has a 3 symbol:
> 
> Image

I removed the icons from menu. It a big topic for another PR

tomk-amd (approved):

drchen-amd (dismissed):
Regression vs dev; Tables 18.3 and 18.4 are mashed together.

https://github.com/user-attachments/assets/0365c785-b46b-4812-b114-4c602bc49b26

amokiche-amd:
> Regression vs dev; Tables 18.3 and 18.4 are mashed together.
> 
>  Screencast.from.04-22-2026.11.44.06.AM.webm



> Regression vs dev; Tables 18.3 and 18.4 are mashed together.
> 
>  Screencast.from.04-22-2026.11.44.06.AM.webm

I didn't know we could have tables with the same title. Okay, fixed. I think this PR is cursed.

src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
```diff
@@ -727,6 +699,74 @@ KernelMetricTable::Render()
     m_query_builder.Render();
 }
 
+void
+KernelMetricTable::SetQuery(const std::string& query)
+{
+    // only add metric if not already added
+    if(std::find(m_metrics_params.begin(), m_metrics_params.end(), query) !=
```
drchen-amd, line 706: This is a preexisting issue but I will mention it here anyway. 
This  is not adequate to prevent ID conflict because two different metrics with the same name can pass. For example here I added two of the many metrics named `0` found section 18.
Image

src/view/src/icons/rocprovfis_icon_defines.h
```diff
@@ -27,32 +27,41 @@ constexpr ImWchar icon_ranges[] = {
     0xF454, 0xF454,
     0xF472, 0xF472,
     0xF484, 0xF484, 
+    0xF127, 0xF127,
```
drchen-amd: This file should be reverted. None of the additions are being used.

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,418 +3,563 @@
...
 #include "rocprofvis_compute_widget.h"
 #include "compute/rocprofvis_compute_selection.h"
+#include "rocprofvis_event_manager.h"
+#include "rocprofvis_events.h"
 #include "rocprofvis_data_provider.h"
 #include "rocprofvis_gui_helpers.h"
 #include "rocprofvis_requests.h"
 #include "rocprofvis_settings_manager.h"
 #include "widgets/rocprofvis_notification_manager.h"
+#include "icons/rocprovfis_icon_defines.h"
```
drchen-amd: Not needed.

src/view/src/compute/rocprofvis_compute_kernel_details.h
```diff
@@ -49,6 +49,7 @@ class ComputeKernelDetailsView : public RocWidget
     EventManager::SubscriptionToken m_kernel_selection_changed_token;
     EventManager::SubscriptionToken m_metrics_fetched_token;
     EventManager::SubscriptionToken m_new_table_data_token;
+    EventManager::SubscriptionToken m_metric_navigation_token;
```
drchen-amd: This is a odd name considering the event called is called `kComputeShowMetricInKernelDetails`. Seems to me like no "navigation" is happening.

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,418 +3,563 @@
...
-            }
+            RenderEmptyTable();
+            return;
         }
-        else
+    }
+
+    const std::string child_id = "##" + m_table_title + "_layout_" +
+                                 std::to_string(reinterpret_cast<uintptr_t>(this));
+    const std::string table_id = child_id + "_table";
```
drchen-amd: Can just use RocWidget::GenUniqueName()

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -3,418 +3,563 @@
...
+
+    if(ImGui::BeginChild(child_id.c_str(), ImVec2(0, GetTableHight()), false,
```
drchen-amd: All this window does is contain the table and allow scrolling. Instead, the table itself should be allowed to scroll. This would also solve the problem of row/column headers being scrolled off screen.

drchen-amd (commented):
Functionality seems fine. I'm not not going to request changes and leave  comments in the interest of time.

Having used this feature for a while, I see some usability issues:

- The column and row headers can be scrolled off screen. This makes it difficult to  know what your actually looking at.Image

- Being taken back to the first page when switching workloads is annoying.

amokiche-amd:
> Functionality seems fine. I'm not not going to request changes and leave comments in the interest of time.
> 
> Having used this feature for a while, I see some usability issues:
> 
> * The column and row headers can be scrolled off screen. This makes it difficult to  know what your actually looking at.Image
> * Being taken back to the first page when switching workloads is annoying.

Thanks good think that we can implement as a second improvment, I totaly agree with that espetialy with case when user scroll down far, but this PR already have a lot of changes, I want merge it asap

src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
amokiche-amd: Good point, but it not regression for this PR, we deffenetly should fix it on the next PR

src/view/src/icons/rocprovfis_icon_defines.h
amokiche-amd: I removed the unused icons, but to do not forget in the future preferer to add inline for the icons.

src/view/src/widgets/rocprofvis_compute_widget.cpp
amokiche-amd: I'd like to preserve the ability for regular metric tables to fill all available space, since, unlike comparison tables, they don't have as many data columns, and it looks much better if they can fill all available space. At the same time, PinnedMetricTable should be able to scroll horizontally. I'm willing to consider this issue, but it's not part of this PR campaign.

drchen-amd:
Already approved. Feel free to dismiss my comments.

PR #768, opened 2026-04-10
Add support for workload metrics to view's compute model and add SOL to summary view

tomk-amd:
## Motivation

Add SOL table to Compute Summary view.

## Technical Details

Add support for workload metrics to compute model,
 - renamed metric getters from generic names to specific names ex: GetMetricValue() -> GetKernelMetricValue()
 - added getters for workload metrics

Add a WorkloadMetricTableWidget class for rendering workload metric tables.

Add speed of light workload metric table to summary view.


src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -258,7 +258,53 @@ MetricTableWidget::UpdateTable()
...
+    DataProvider& data_provider, std::shared_ptr<ComputeSelection> compute_selection,
+    uint32_t category_id, uint32_t table_id)
+: MetricTableWidget(data_provider, compute_selection, category_id, table_id)
+{}
+
+void
+WorkloadMetricTableWidget::FetchMetrics()
+{
+    m_table.Clear();
+    m_data_provider.ComputeModel().ClearKernelMetricValues(m_client_id);
```
drchen-amd: This should clear the workload store.

drchen-amd (changes requested):

src/view/src/widgets/rocprofvis_compute_widget.cpp
tomk-amd: good catch - it worked because this store id isn't shared

drchen-amd (approved):

PR #773, opened 2026-04-14
Compute Comparison V2

drchen-amd:
- Selective coloring for metrics with non-zero differences.
  - Updated `ComputeComparisonView::Table` to allow each cell to carry visual customizations:
    - Text color, BG color, icon. 
  - Removed rule-based column coloring.
- Fixed precision errors in difference values by rounding prior to calculation.
  image

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_comparison.cpp
+++ b/src/view/src/compute/rocprofvis_compute_comparison.cpp
@@ -406,7 +406,7 @@ ComputeComparisonView::UpdateMetrics()
                                             std::min(static_cast<ImU32>(
                                                          std::abs(rounded_target -
                                                                   rounded_baseline) /
-                                                         rounded_baseline * 255),
+                                                         rounded_baseline * 255 + 25),
                                                      ImU32(255));
                                     }
                                 }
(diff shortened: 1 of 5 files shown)


src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -361,65 +360,135 @@ ComputeComparisonView::UpdateMetrics()
...
+                                    }
+                                    if(rounded_baseline != 0.0)
+                                    {
+                                        bg_color_difference.value().alpha =
+                                            std::min(static_cast<ImU32>(
+                                                         std::abs(rounded_target -
+                                                                  rounded_baseline) /
+                                                         rounded_baseline * 255),
+                                                     ImU32(255));
+                                    }
```
tomk-amd, line 412: I would make sure a minimum alpha greater than some threshold is always set if there is a difference.  Without this small changes in low single digit percentages or below are either imperceivable or just have 0 alpha, for example always have an alpha of at least 25 if there is a difference detected:


```suggestion
                                    if(rounded_baseline != 0.0)
                                    {
                                        bg_color_difference.value().alpha =
                                            std::min(static_cast<ImU32>(
                                                         std::abs(rounded_target -
                                                                  rounded_baseline) /
                                                         rounded_baseline * 225 + 25),
                                                     ImU32(255));
                                    }
```

tomk-amd (approved):
Other than the miminum alpha suggestion, looks good.

PR #776, opened 2026-04-15
Measure Mode

Sajeeth-Wimalasuriyan:
image

**Summary**
Adds a Measurement Mode to the timeline view that lets users measure the time delta between two points on the timeline.

Two modes:
Event mode — click two flame events to snap rulers to their start/end edges (toggleable per point).
Freehand mode — click anywhere on the timeline to drop rulers, then drag them horizontally to fine-tune.

Visuals: two vertical rulers with timestamp labels, a connector line between them, event-name labels, and a delta label at the midpoint. Theme-aware colors for both dark and light themes.

Controls: new toolbar group (Measure On/Off, per-point Start/End edge toggles, Freehand toggle) plus matching entries in the timeline right-click menu.

Highlighting: measured events stay highlighted via a new persistent-highlight path. Refactored TimelineSelection from a single global highlight timer to per-event timers so persistent (measurement) and transient (search/navigate) highlights coexist cleanly.


amokiche-amd:
Could you please fill out a brief description of the request? It's not obvious what you did here without context, unless you dig into the code.

drchen-amd:
Blind first impression:
- The non-free hand mode needs better tool tips. I initially tried clicking and dragging around which did nothing, then I thought it might be keyboard activated and looked in the hotkey menu but did not see any keys. User should be informed that they need to click two events.
- Feels somewhat clunky that after picking your start and end, your clicks do nothing until you clear. Maybe it could not limit to one range at a time and allow multiple ranges where every odd click drops down a start, and every even click drops down an end.
- Context menu says "clear measurement", tool bar says "reset". They should both be either "clear" or "reset" assuming they do the same thing.
- Maybe move bookmarks to the left side of search bar. I think appearance wise it would be better if search bar is at least mostly centered.

vstempen:
I would rather use context menu to "measure from here", "measure to here", almost like Google does, another set of options "measure from this event", "measure to this event". I think it's simpler and more intuitive.

amokiche-amd:
1.When you enter to measure mode chose two events and then exit from measure mode by context menu the selection of mesurement doesn't removed
image
2. I guess for user will be useful to reselect the events for measurement not only by reset button. For example second click for already chosen event should unselect it and allow user chose another one.
3. nit: This is purely my preference, but what if I'm not alone. I find that buttons with status that change state are a bit confusing for the user. Sometimes it's not clear to me what exactly it displays. When I see "measure mode off" on a button, I always wonder if it's the status off or if I'll  turn it off when I press it. But again, maybe I'm the only one who always wonders about this.

amokiche-amd:
1. When a user switches to the Anywhere mode, the event selection should be reset, since there is no point in event selection in such a mod. 
image

2. nit: I'm not insisting on this, but I think this workflow is a little more intuitive. If it's not too difficult to rework, of course. Everything is fine if two events are selected and the user clicks the third event, then we start the selection from the beginning. But I think it would be more convenient if the user clicks on an already selected event, then this cancels its selection, rather than starting the selection process from scratch. Example: Events 123 and 456 are selected, the user clicks event 123 a second time. In this implementation, event 123 will be selected, and 456 will be reset. It would be more convenient if in this case the selection state changed to the event the user clicked, namely 123.

3. one more nit :) : The measurement button is on the right, and the measurement mode interface appears on the left. This is a bit confusing. I think it would be neater to either move the measurement mode button to the left or align the measurement mode interface to the right. But I'm not sure; I'd like to hear other opinions on this matter.

4. I don't know may it useful or not. But looks like not intentional. I guess we should unselect events and hide the mesure ruler when user exit from measure mode
image

src/view/src/rocprofvis_click_manager.h
```diff
@@ -32,6 +60,28 @@ class TimelineFocusManager
...
+
+    const MeasurementPoint& GetPoint(int index) const;
+    MeasureEdge             GetEdge(int index) const;
+    void                    SetEdge(int index, MeasureEdge edge);
+    double                  GetEffectiveTimestamp(int index) const;
+
+    void   SetFreehandMode(bool enabled);
+    bool   IsFreehandMode() const;
+    void   SetFreehandOffset(int index, double offset_ns);
+    double GetFreehandOffset(int index) const;
```
tomk-amd: I don't think that these functions should be in TimelineFocusManager.   A dedicated Measurement class should be created and owned by timeline.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -219,6 +221,121 @@ TimelineView::RenderAnnotations(ImDrawList* draw_list, ImVec2 window_position)
...
+
+    // Draws a small timestamp label centered on a ruler line
+    auto draw_ruler_label = [&](float x, const char* text) {
+        ImVec2 sz = ImGui::CalcTextSize(text);
+        float  lx = x - sz.x * 0.5f;
+        draw_list->AddRectFilled(ImVec2(lx - 4.0f, label_y - 2.0f),
+                                 ImVec2(lx + sz.x + 4.0f, label_y + sz.y + 2.0f),
+                                 label_bg, 3.0f);
+        draw_list->AddText(ImVec2(lx, label_y), label_text, text);
+    };
```
tomk-amd, line 289: lot's of magic numbers here? should it be lx - LABEL_PAD  ?

src/view/src/rocprofvis_click_manager.h
tomk-amd: In the interest of getting this merged in, this can be refactored in a separate PR.

tomk-amd:
I think this works well now, except for a small nitpik:

image

The spacing for the mode button is inconsistent, there should be a gap between "Mode"  and the button.

**And one major issue:** 

Measure mode is maintained between traces (tabs), not sure if this is correct or not.   But what is worse is that the measurement is shared between tabs - traces.   I take back the comment about moving the logic to a separate class owned by timeline via a new PR.   This has be done now, the measurement data / operations have to be tied to each timeline not the global focus manager.

tomk-amd (changes requested):
Measurement data can't exist in global object.  Must be owned by each timeline (trace).

tomk-amd (approved):

PR #777, opened 2026-04-15
Random Compute View Fixes

tomk-amd:
## Motivation

Fix a number of aesthic items.
 
## Technical Details

- Unified UX look and feel for showing / hiding tables (pinned metrics - formerly bookmarks on comparison view, and kernel selection table on kernel details view)
- Renamed bookmarks to pinned metrics in both code and displayed text in comparison view.
- Fix kernel name tooltip behaviour.
- Fast DB cleanup option hidden from end users.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_comparison.cpp
+++ b/src/view/src/compute/rocprofvis_compute_comparison.cpp
@@ -49,21 +49,21 @@ ComputeComparisonView::ComputeComparisonView(
 , m_layout(nullptr)
 , m_toolbar_available_width(0.0f)
 , m_tab_container(nullptr)
-, m_bookmark_table(nullptr)
-, m_bookmark_item(nullptr)
-, m_max_bookmark_height(FLT_MAX)
+, m_pinned_table(nullptr)
+, m_pinned_item(nullptr)
+, m_max_pinned_height(FLT_MAX)
 {
     m_widget_name = GenUniqueName("ComputeComparison");
-    m_bookmark_table =
+    m_pinned_table =
         std::make_unique<Table>("",
                                 ImGuiTableFlags_Borders | ImGuiTableFlags_RowBg |
                                     ImGuiTableFlags_ScrollX | ImGuiTableFlags_ScrollY,
                                 3, 1);
-    m_bookmark_table->SetRowSelectionHandler(
+    m_pinned_table->SetRowSelectionHandler(
         [this](const Table& table, const size_t index, const bool state) {
             if(!state)
             {
-                RemoveBookmark(table, index);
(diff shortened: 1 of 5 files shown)


src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -744,18 +754,18 @@ ComputeComparisonView::RenderCategory(const size_t i)
...
             ImVec2(ImGui::GetContentRegionAvail().x, ImGui::GetFrameHeightWithSpacing()),
-            ImVec2(ImGui::GetContentRegionAvail().x, m_max_bookmark_height));
+            ImVec2(ImGui::GetContentRegionAvail().x, m_max_pinned_height));
         ImGui::BeginChild("bookmarks", ImVec2(0.0f, 0.0f),
                           ImGuiChildFlags_Borders | ImGuiChildFlags_AutoResizeY);
-        if(m_bookmarks.empty())
+        if(m_pinned_metrics.empty())
         {
-            CenterNextItem(ImGui::CalcTextSize("Use () to bookmark metrics.").x +
+            CenterNextItem(ImGui::CalcTextSize("Use () checkbox to pin metrics.").x +
```
drchen-amd, line 768: I'm not sure about adding "checkbox" here because now seems like checkbox is there twice.

"Use [checkbox] checkbox to pin metrics"

src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -700,36 +695,51 @@ ComputeComparisonView::RenderToolbar()
...
+
+        ImGui::PushFont(icon_font);
+        // use larger icon for consistent spacing
+        ImVec2 icon_size = ImGui::CalcTextSize(ICON_CHEVRON_DOWN); 
+        ImGui::PopFont();
+
+        if(IconButton(icon, icon_font,
+                      ImVec2(icon_size.x + style.FramePadding.x * 2.0f,
+                             icon_size.y + style.FramePadding.y * 2.0f),
+                      m_pinned_item->m_visible ? "Hide pinned metrics" : "Show pinned metrics",
```
drchen-amd, line 720: I think it makes more sense if the icon was on the right side of the label.
Also, this icon button helper was never updated to use the rounded tooltips.

drchen-amd (approved):

src/view/src/compute/rocprofvis_compute_comparison.cpp
tomk-amd: switched order - i think you're right, it looks better.

Sajeeth-Wimalasuriyan (approved):
We are good to go

PR #778, opened 2026-04-16
Comparison view delta threshold control

tomk-amd:
## Motivation

Add a configurable delta-threshold control for compute comparison metrics.  Allows user to suppress the noise of minor deviations or filter out .changes below a certain level.

## Technical Details

Add metadata fields to Value Structs (used for locating / matching cells when updating bg colors):
            Role                  role = Role::Other;
            std::string           compare_group;
Add control to toolbar, call UpdateDifferenceHighlighting() on value change.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_comparison.cpp
+++ b/src/view/src/compute/rocprofvis_compute_comparison.cpp
@@ -123,7 +123,6 @@ ComputeComparisonView::Update()
         {
             m_tab_container->SetActiveTab(m_active_tab_id);
         }
-        m_data_changed = false;
     }
     for(const CategoryModel& category : m_categories)
     {
@@ -138,6 +137,11 @@ ComputeComparisonView::Update()
     if(m_pinned_table)
     {
         m_pinned_table->Update();
+    }
+    if(m_data_changed)
+    {
+        UpdateDifferenceHighlighting();
+        m_data_changed = false;
     }
 }
 
@@ -283,6 +287,7 @@ ComputeComparisonView::UpdateMetrics()
                                                         m_target_kernel_id))
         {
             m_categories.clear();
+            m_diff_cell_groups.clear();
             m_tab_container = std::make_unique<TabContainer>();
             std::vector<const AvailableMetrics::Category*> baseline_categories =
                 m_data_provider.ComputeModel()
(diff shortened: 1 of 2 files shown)


src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -528,6 +533,179 @@ ComputeComparisonView::UpdateMetrics()
...
+                diff_value.display_props.icon         = std::nullopt;
+                diff_pct_value.display_props.icon     = std::nullopt;
+
+                if(baseline_value.data.empty() || target_value.data.empty())
+                {
+                    continue;
+                }
+
+                const double rounded_baseline = std::atof(baseline_value.data.c_str());
+                const double rounded_target   = std::atof(target_value.data.c_str());
```
drchen-amd: I don't have a strong opinion about this but maybe each `Table::Row::Value` can store the original double in `Table::AddRow` instead of this conversion.

Also, for this color update only path, I don't think it is necessary to compute the difference. It should be sufficient to read the `diff_pct_value` and compare with the threshold.

src/view/src/compute/rocprofvis_compute_comparison.h
```diff
@@ -104,6 +116,7 @@ class ComputeComparisonView : public RocWidget
 
         // Getters...
         const std::vector<Row>&         Rows() const;
+        std::vector<Row>&               MutableRows();
```
drchen-amd: I think it would be safer if instead of adding this, the `Table::Row::Value`'s `display_props` is made mutable, like how `selection` is.

src/view/src/compute/rocprofvis_compute_comparison.h
```diff
@@ -53,9 +53,19 @@ class ComputeComparisonView : public RocWidget
         // Clients populate data via Rows of Values...
         struct Value
         {
+            enum class Role
```
drchen-amd: It would be nice if `Table` continues to be generic and free from concepts of `ComparisonView` like baseline and target...etc.

I believe one way to accomplish this is to have `ComparisonView` keep a list of references to the `Difference (%)` cells's `Row::Value` as rows are added. Then the threshold coloring can be updated by iterating through the list, comparing the `Row::Value.data` with the threshold, and changing the `Row::Value::DisplayProp`.

tomk-amd: I've taken an alternate approach here:  https://github.com/ROCm/roc-optiq/compare/dev...tkarczew/delta_threshold_v2

If the intent of the Table class was / is to be general then it should live outside of  ComputeComparisonView or at very least not be private

drchen-amd: Yes the intention was to keep it private until there was a second user, but have the implementation be general such that the class could just be copied and pasted out when the time comes.

src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -137,6 +139,16 @@ ComputeComparisonView::Update()
     if(m_pinned_table)
     {
         m_pinned_table->Update();
+        if(m_pinned_columns_dirty)
+        {
+            RebuildPinnedDiffEntries();
+            m_pinned_columns_dirty = false;
+        }
+    }
+    if(m_data_changed)
```
drchen-amd, line 148: I would prefer if UpdateDifferenceHighlighting() was split into pinned and non pinned functions just so the m_data_changed_ handler is not split into two parts like this.

Feel free to leave it as is. If you agree, I can make the change since I will need to work with pins anyway for the presets.

drchen-amd (approved):

src/view/src/compute/rocprofvis_compute_comparison.cpp
tomk-amd: The main issue for the split is that UpdateDifferenceHighlighting() needs to be called after this block runs..

```
    for(const CategoryModel& category : m_categories)
    {
        for(const std::shared_ptr<Table>& table : category.tables)
        {
            if(table)
            {
                table->Update();
            }
        }
    }
```

Since update metrics changes the table(s).  I'd prefer to leave this refactoring to the presets PR since you're more familiar with this part of the code.

PR #782, opened 2026-04-17
Filter out non-finite metric values in kernel selection table

Sajeeth-Wimalasuriyan:
Cells parsing to inf/-inf/NaN are now displayed as "N/A" and excluded from bar chart max-value computation, so a single infinite value no longer collapses bar scaling for other rows in the same column.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
+++ b/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
@@ -44,6 +44,20 @@ constexpr float COL_INVOCATION_CHAR_LIMIT = COL_FILTER_CHAR_LIMIT;
 
 constexpr float kTooltipMaxWidth = 600.0f;
 
+namespace
+{
+// Returns true if `cell` parses as a numeric value that is not finite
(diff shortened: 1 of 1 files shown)


src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
```diff
@@ -592,8 +606,11 @@ KernelMetricTable::Render()
                             }
                             else
                             {
-                                if(cell.empty())
+                                if(cell.empty() || IsNonFiniteNumericCell(cell))
                                 {
+                                    // Treat inf / -inf / NaN as unavailable so
+                                    // they don't clutter the table or skew the
+                                    // bar chart scaling.
```
tomk-amd: I don't think we need to filter out the displayed value - I think it is ok to leave it as it is.  

Might even be helpful as in one of the samples the infinity was coming directly from the database, so it will allow us to spot these issues.

But more importantly this string -> double conversion ends up being done for every cell and every frame.  If we do decide to filter out these values it should be done when the data changes.

Please remove IsNonFiniteNumericCell() and this check

tomk-amd (approved):

PR #789, opened 2026-04-21
Prevent accidental Roofline zoom while scrolling

tomk-amd:
## Motivation

When using the mouse wheel to scroll the view / page once the mouse hovers over the roofline chart the scrolling is stopped and instead the chart begins to zoom in.  This makes for an awkward user experience. 

Instead the chart now will need to be "focused" by being clicked before scroll wheel (and other) interaction is allowed.

Pressing Esc or clicking off of the chart will exit the interactive state.

## Technical Details

Add flag to enable/disable plot interactions. (Click on chart to enable interaction) .

Add zoom hint for roofline chart when hovered.

Clean up compiler warnings.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_roofline.cpp
+++ b/src/view/src/compute/rocprofvis_compute_roofline.cpp
@@ -555,11 +555,6 @@ Roofline::Render()
         bool roofline_hovered = plot_size.x > 0.0f && plot_size.y > 0.0f &&
                                 ImGui::IsMouseHoveringRect(
                                     plot_pos, plot_pos + plot_size, false);
-        if(m_plot_interactions_enabled && roofline_hovered)
-        {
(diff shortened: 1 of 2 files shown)


drchen-amd:
I can see the motivation for disabing zoom, but I don't see the harm in having the other interactions (hover, crosshair) being always on. I think it would be better if the other interactions were unaffected. Locking the axis using using `ImPlotAxisFlags_Lock` can acomplish this.

Otherwise for current implementation, the text is not accurate because more than just zoom is affected.

tomk-amd:
I agree that ideally other hover operations should not be impacted, but `ImPlotAxisFlags_Lock` does not seem to work.  With this flag enabled the chart is still zoomable.  I can change the text to "Click to interact with chart".

*Update* :  `ImPlotAxisFlags_Lock`  works, my mistake, I was setting this in the wrong call.

drchen-amd (approved):

PR #793, opened 2026-04-22
Track Cleanup

Sajeeth-Wimalasuriyan:
Cleaned up the track sidebar UX in a few small ways: bumped the default track height to fit two event rows (so 2-event tracks no longer prompt to expand), replaced the gear button with a right-click context menu on the metadata area, dimmed the reorder grip to kTextDim grey, dropped the "Color by Time Level" mode entirely along with the "No Color" / "Color by Name" radios so the track options menu is just Compact Mode now.

### DIFF
--- a/src/view/src/rocprofvis_flame_track_item.cpp
+++ b/src/view/src/rocprofvis_flame_track_item.cpp
@@ -110,6 +110,18 @@ FlameTrackItem::FlameTrackItem(DataProvider&                      dp,
 void
 FlameTrackItem::RenderMetaAreaExpand()
 {
+    int visible_levels =
+        static_cast<int>(std::floor(m_track_content_height / m_level_height));
+
+    const bool show_expand   = visible_levels < m_max_level + 1;
+    const bool show_contract = !show_expand &&
+                               m_track_height > std::max(m_max_level * m_level_height +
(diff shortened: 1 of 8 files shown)


src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -110,6 +110,23 @@ FlameTrackItem::FlameTrackItem(DataProvider&                      dp,
...
+    //
+    // The original code (Sept 2025) hit this any time a single-event flame track sat
+    // at default height. Bumping DEFAULT_TRACK_HEIGHT and tightening the visibility
+    // check earlier in this branch made 2-event tracks at default height fall into
+    // the same "no arrow" path too, which is the common case, so the assert started
+    // firing constantly (especially while dragging since every frame redraws every
+    // visible track).
+    //
+    // Fix: figure out which arrow we're going to draw up front and return early when
+    // there's nothing to render so we don't touch the cursor at all.
```
tomk-amd: This can go as PR comment, it shouldn't be in the code

tomk-amd:
After expanding track with tall flame chart the "collapse" arrow does not appear:

image

Need to manually adjust track height a bit taller to make it show again

PR #794, opened 2026-04-22
File dialog fallback

tomk-amd:
## Motivation

Improvements to the file dialog infrastructure are needed.  A number of issues exist with the current implementation such as:
- native dialog not being forwarded over remote ssh sessions
- application crashing if dbus is not present or native file dialog fails to initialize for some other reason 

## Technical Details

Runtime file dialog fallback for SSH sessions: Both the native (xdg-desktop-portal / macOS) file dialog and the built-in ImGui file dialog are now compiled into every build. 

At startup the app auto-detects remote sessions (SSH_CONNECTION, SSH_CLIENT, SSH_TTY, or forwarded DISPLAY) and transparently uses the ImGui dialog, which forwards correctly over ssh -X.

Graceful NFD_Init failure handling: NFD_Init() is probed at startup and checked again at each dialog open. If it fails (e.g. no D-Bus), the app permanently downgrades to the ImGui dialog instead of crashing.

--file-dialog CLI flag: New `--file-dialog=auto|native|imgui option (-d)`  to override auto-detection.

macOS native dialog support (merged from sw-macos-file-explorer): Removes the blanket macOS USE_NATIVE_FILE_DIALOG=OFF override. The native dialog now runs synchronously on the main thread via std::promise to satisfy AppKit's threading requirements.

## Test Plan

Build with
USE_NATIVE_FILE_DIALOG=ON
(default) on Linux — verify both dialog paths work

Build with
USE_NATIVE_FILE_DIALOG=OFF
— verify ImGui-only path compiles and runs

Run over
ssh -X
— verify auto-detection picks ImGui and file dialog appears on the client

Run
--file-dialog=imgui
locally — verify override works

Run
--file-dialog=native
over SSH — verify native is forced

Run on a system without D-Bus/portal — verify graceful fallback, no crash

Verify macOS native file dialog opens correctly (main-thread dispatch)

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -203,3 +203,33 @@ cmake --build build/macos-release --preset "macOS Release Build" --parallel 4
 - macOS: the executable is in `build/<preset>/`.
 
 If you need symbol builds, use the `*-release-symbols` presets.
+
+---
+
+## File dialog behavior on Linux
+
+On Linux, `roc-optiq` uses the native [xdg-desktop-portal](https://flatpak.github.io/xdg-desktop-portal/) file chooser by default, but also ships the in-process ImGui file dialog as a fallback.
+
+The portal dialog is launched by an external D-Bus service on the host machine and is **not** compatible with remote display forwarding (e.g. `ssh -X` / `ssh -Y`): the portal parents its window on the host's compositor, so over an SSH session the dialog would appear on the host rather than on your client — or simply never appear at all if the host has no local display.
+
+To work around this, the application automatically picks the in-window ImGui dialog when it detects a remote session. Detection looks at:
+
+- `SSH_CONNECTION`, `SSH_CLIENT`, `SSH_TTY` environment variables (set by `sshd`), and
+- `DISPLAY` matching `localhost:N` with `N >= 10` (the range SSH uses for X11 forwarding).
+
+The chosen backend is logged once at startup. If the auto-detection gets it wrong (for example the SSH environment was stripped by `sudo` or `systemd-run`, or you are `ssh`'ing into the same host that also runs a local desktop), you can override it with the `--file-dialog` command-line flag:
+
+```bash
+# Force the in-process ImGui dialog regardless of detection
+roc-optiq --file-dialog=imgui
+
+# Force the native (xdg-desktop-portal) dialog
+roc-optiq --file-dialog=native
+
+# Let the app auto-detect (the default)
+roc-optiq --file-dialog=auto
+```
+
+If `xdg-desktop-portal` or D-Bus is not available on the host, the native dialog probe will fail gracefully at startup (or at the first dialog open) and the app will automatically fall back to the ImGui dialog for the remainder of the session.
+
+If you want to disable the native dialog entirely at build time (so ImGui is always used), configure with `-DUSE_NATIVE_FILE_DIALOG=OFF`.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,9 +42,6 @@ endif(ROCPROFVIS_DEVELOPER_MODE)
 if(ROCPROFVIS_ENABLE_INTERNAL_BANNER)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DROCPROFVIS_ENABLE_INTERNAL_BANNER")
 endif(ROCPROFVIS_ENABLE_INTERNAL_BANNER)
-if(USE_NATIVE_FILE_DIALOG)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_NATIVE_FILE_DIALOG")
-endif(USE_NATIVE_FILE_DIALOG)
 
 # Set GLFW library type based on option
 if(GLFW_LINK_STATIC)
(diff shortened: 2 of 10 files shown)


drchen-amd (commented):
Confirmed the ssh auto fallback works with 22.04.
Is it expected that I had to manually specify opengl flag with ssh? Vulkan would crash.
(I tested by ssh into localhost.)

tomk-amd:
> Confirmed the ssh auto fallback works with 22.04. Is it expected that I had to manually specify opengl flag with ssh? Vulkan would crash. (I tested by ssh into localhost.)

No crash is not expected, and not related to this PR..  I saw that when testing remote forwarding with 22.04 too. (It happens on dev).. I haven't had a chance to debug it yet to see what is causing it.

tomk-amd:
For posterity's sake:

```
[2026-04-23 10:35:56.358] [rocprofvis-log] [info] [rpv] Vulkan is supported, attempting Vulkan backend...                                                                     
[2026-04-23 10:35:56.358] [rocprofvis-log] [info] [rpv] Using Vulkan backend
warning: could not find '.gnu_debugaltlink' file for /lib/x86_64-linux-gnu/libVkLayer_MESA_device_select.so
Downloading separate debug info for /lib/x86_64-linux-gnu/libVkLayer_MESA_device_select.so
[New Thread 0x7fffd47ff6c0 (LWP 1862446)]                                                                                                                                     
[New Thread 0x7fffd3ffe6c0 (LWP 1862447)]
[New Thread 0x7fffd36bc6c0 (LWP 1862448)]
[New Thread 0x7fffd2d7a6c0 (LWP 1862449)]

Thread 1 "roc-optiq" received signal SIGSEGV, Segmentation fault.
0x00007fffd559938b in ?? () from /lib/x86_64-linux-gnu/libvulkan_radeon.so
(gdb) bt
#0  0x00007fffd559938b in ??? () at /lib/x86_64-linux-gnu/libvulkan_radeon.so
#1  0x0000555555917d6a in ImGui_ImplVulkanH_CreateWindowSwapChain
    (physical_device=0x5555566709a0, device=0x5555566a3020, wd=0x55555655f0b0, allocator=0x0, w=1280, h=720, min_image_count=2, image_usage=0)
    at /home/rocm/projects/rocprofiler-visualizer/thirdparty/imgui/backends/imgui_impl_vulkan.cpp:1725
#2  0x00005555559184df in ImGui_ImplVulkanH_CreateOrResizeWindow
    (instance=0x55555662a870, physical_device=0x5555566709a0, device=0x5555566a3020, wd=0x55555655f0b0, queue_family=0, allocator=0x0, width=1280, height=720, min_image_count=2, image_usage=0) at /home/rocm/projects/rocprofiler-visualizer/thirdparty/imgui/backends/imgui_impl_vulkan.cpp:1831
#3  0x00005555555aae44 in rocprofvis_imgui_backend_vk_init (backend=0x7fffffffdf10, window=0x55555653a990)
    at /home/rocm/projects/rocprofiler-visualizer/src/app/src/rocprofvis_imgui_vulkan.cpp:451
#4  0x00005555555a90cc in rocprofvis_imgui_backend_complete_init_with_opengl_fallback
    (backend=0x7fffffffdf10, window=0x7fffffffde78, width=1280, height=720, title=0x55555614901a "ROCm(TM) Optiq Beta", preference=kRPVBackendAuto)
    at /home/rocm/projects/rocprofiler-visualizer/src/app/src/rocprofvis_imgui_backend.cpp:145
#5  0x00005555555b04c2 in main (argc=1, argv=0x7fffffffe268) at /home/rocm/projects/rocprofiler-visualizer/src/app/src/main.cpp:262
(gdb) 
```

AI Analysis

Here's what's happening:

Root cause: The crash is inside libvulkan_radeon.so at the vkGetSwapchainImagesKHR call (line 1725 of imgui_impl_vulkan.cpp). Over SSH X11 forwarding, the RADV (Radeon Vulkan) driver manages to:

Create a Vulkan instance and physical device (these don't need a real display)
Create a surface via GLFW (succeeds)
Report that the physical device supports presentation on that surface (vkGetPhysicalDeviceSurfaceSupportKHR returns VK_TRUE)
Even vkCreateSwapchainKHR at line 1723 appears to return VK_SUCCESS (otherwise check_vk_result would have fired)
But then when vkGetSwapchainImagesKHR tries to actually query the swapchain images, the driver segfaults instead of returning an error code. This is a bug in libvulkan_radeon.so - it should return a VK_ERROR_* code, not crash.

Why the fallback doesn't work: The app has a nice OpenGL fallback in rocprofvis_imgui_backend_complete_init_with_opengl_fallback, but it only catches cases where backend->m_init() returns false. A SIGSEGV kills the process before the fallback logic ever runs.

The practical fix is to detect SSH X forwarding and skip Vulkan. Let me show you the simplest approach - adding SSH detection in the backend selection logic:

Here's my analysis:

The problem: The Radeon Vulkan driver (libvulkan_radeon.so) has a bug where it segfaults inside vkGetSwapchainImagesKHR (or possibly vkCreateSwapchainKHR) when running over SSH X11 forwarding. It reports that it supports presentation on the surface, but then crashes instead of returning an error code when actually trying to use the swapchain. Since this is a hard crash (SIGSEGV), the app's OpenGL fallback in rocprofvis_imgui_backend_complete_init_with_opengl_fallback never gets a chance to run

drchen-amd (approved):

PR #795, opened 2026-04-23
Sidebar Cleanup

Sajeeth-Wimalasuriyan:
Replace the per-track eye and scroll-to-track icons with a Selectable that uses single-click to toggle selection and double-click to scroll to the track. Visibility is now driven by a right-click context menu on both leaf rows (Show/Hide Track) and branch headers (Show/Hide All), keeping the existing bulk-toggle and cache machinery. Hidden tracks remain greyed out via TextDisabled.

Tighten and modernize the layout: bump indent and padding, soften the tree-connector lines (thinner stroke, ~50% alpha), and lengthen the elbow stub so it lands cleanly at each row.

### DIFF
--- a/src/view/src/rocprofvis_sidebar.cpp
+++ b/src/view/src/rocprofvis_sidebar.cpp
@@ -198,6 +198,11 @@ SideBar::RenderTrackItem(const uint64_t& index, bool allow_visibility_toggle)
                                    graph.display ? hidden_chart_ids : shown_chart_ids);
                 UpdateHistogramForVisibility(shown_chart_ids, hidden_chart_ids);
             }
+            if(ImGui::MenuItem("Show All Tracks", nullptr, false,
+                               HasTrackVisibility(false)))
+            {
+                ApplyAllTrackVisibility(true);
+            }
             if(ImGui::MenuItem("Hide All But This Track"))
             {
                 HideAllButTrack(index);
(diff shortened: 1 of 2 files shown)


tomk-amd:
The context menus need to have padding.

I think the context menu option text needs to be reworked for parent nodes:
- hide all tracks in this group or hide all tracks below?
- show all tracks in this group or show all tracks below?

For track nodes the following context menu options should be added:
- go to track (I know double click does this, but ... it is ok to have another way)
new features:
- hide all but this track (user requested this feature) 
- hide / show selected tracks

It would be nice if there was an icon beside the context menu action. (This can be done after the font system issues are resolved).

amokiche-amd:
1. Context menu doesn't work for root of tree, if i clicked on "Project" nothing happens.
2. Tom wrote above that a double click should focus on a track, but I see you disabled it? I guess it will be usefull to have the both ways for navigate to track.
3. Regarding the icons in the context menu, I will check whether the icons work properly on all platforms.
4. I’d also rename “Hide All But This Track” to “Show Only This Track”. It is shorter and easier to understand.
5. Discussion question: I think the context menu with so many options is a bit noisy. I think it would be useful to add a couple of submenus to make the main menu shorter, something like this option:
image
image

tomk-amd:
> 2. Tom wrote above that a double click should focus on a track, but I see you disabled it? I guess it will be usefull to have the both ways for navigate to track.

We decided to disable "double click" for now as the act of double clicking also causes the track to get selected.   Not obvious how to handle double click with out processing the single click first.. (maybe there is some ImGUI flag / option that can be set), but otherwise we need to write our own click deferring logic.

> Discussion question: I think the context menu with so many options is a bit noisy. I think it would be useful to add a couple of submenus to make the main menu shorter, something like this option:

I agree but not sure I like adding another layer of sub-menu, requires user to make more clicks and mouse movements.  We'll brainstorm some ideas today.

drchen-amd:
I don't think this is an improvement over existing implementation.

Hiding the icon buttons means it is not immediately obvious what operations are supported. Also moving them to context menu adds to the number of clicks required to for common operation.

Of course I'm biased because I added these icons buttons. My inspiration was the layers panel in photoshop and the scene tree in 3dsmax. 

image 
image

PR #797, opened 2026-04-23
Presets

### LINKED ISSUE #772: {Defect] IMGUI Debug Assertion when scrolling outer window in comparison view
Image

IMGUI Debug Assertion crash in comparison view.

Steps to reproduce:

Goto comparison tab select L2 Cache (per channel)
Use outer scrollbar to scroll entire contents (all tables)
Get asserstion

Image

Image

**NOTE** This does not seem to cause any negative effects in Release builds.

drchen-amd:
`PresetManager`

- Static class responsible for reading/writing presets from/to JSON file, signaling `PresentComponent`s to save/load, ensuring `PresentComponent`s do not interfere with each other.
- Maintains single JSON file located app config path, and collection of `PresentComponent`s grouped by project (tab). Signals are only sent to the `PresentComponent`s of the active project.
- Sample: [presets.json](https://github.com/user-attachments/files/27032766/presets.json)

`PresetComponent`
- Abstract class to be implemented by widgets that support presets. Responsible for saving/loading the state of a widget when signaled by `PresetManager`.
- Has a `ComponentType` property. All `PresetComponent` of the same `ComponentType` share the same JSON slot (they overwrite others of the same type). The intention is for this to represent and distinguish between a class of widget. Currently there are three types for the three supported widgets.

`PresetBrowser`
- UI for PresetManager, a small popup. Allows for creating, modifying, loading, and deleting named presets.

`ComputeComparsionView`
- Integrated `PresetComponent` for pinned metrics.
- Changed `PinnedModel` to hold copy of metric identifiers, rather than holding a reference to `ComputeDataModel::AvailableMetrics`. This allows the pins to be loaded from the preset regardless of where the preset was created. 
  - Previously any pin was guaranteed to part of `ComputeDataModel::AvailableMetrics`. With presets that may be created from different DBs, this is no longer guaranteed. 
- Changed unit column to only appear when there is at least one metric with a non-empty unit. 
  - Necessary to not show a column of N/As when loading a preset prior to choosing a comparison target. There are also tables like 3.1 and 18.2 which entirely lack units.
- Simplified difference threshold shading
  - Removed dependency on column indexes. This was done by keeping track of each difference `Table::Row::Cell`'s underlying `Table::Row::Value` inside `Table::Row`'s `value_map` rather than its index in `cells` vector.
  - Only the alpha component is updated when the threshold changes. RGB is set once in `Table::AddRow`.

`KernelMetricTable`
- Integrated `PresetComponent` for selected metrics.
- Fixed metric names becoming mismatched with metric values when switching workload. 
  - This only occurs with specific metrics of DBs containing multiple workloads that have conflicting metric IDs.
  - Ex: image
  - The solution is to update the `MetricInfos` after a workload switch has occurred.

- Omit unit in column header when metric does not have unit information to avoid showing `()`.

`ComputeTableView`
- Integrated `PresetComponent` for pinned metrics.


tomk-amd:
Something has changed in the coloring logic when the percentage move is tiny, ex:

image

should be:

image

drchen-amd:
0% differences were assumed to be "not different" and skipped in processing. Did not consider there could be non-zero differences with 0% difference.

src/view/src/compute/rocprofvis_compute_comparison.cpp
```diff
@@ -1062,26 +960,99 @@ ComputeComparisonView::UpdatePinnedMetrics()
...
+            {
+                group.pct_value->display_props.bg_color.value().alpha = std::min(
+                    static_cast<ImU32>(
+                        std::abs(group.pct_value->dbl_data.value()) / 100.0 * 255 + 25),
+                    ImU32(255));
+                group.value->display_props.bg_color.value().alpha = std::min(
+                    static_cast<ImU32>(
+                        std::abs(group.pct_value->dbl_data.value()) / 100.0 * 255 + 25),
+                    ImU32(255));
+            }
```
tomk-amd, line 1047: Calculate this once?

tomk-amd (approved):

PR #805, opened 2026-04-29
Consistency improvement for metric table view

amokiche-amd:
## Motivation
Consistency improvement for metric table view

## Technical Details
The table header is now fixed when scrolling down.
The selected table is now remembered when changing workloads.

## Submission Checklist

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_table_view.cpp
+++ b/src/view/src/compute/rocprofvis_compute_table_view.cpp
@@ -264,7 +264,6 @@ ComputeTableView::RenderCategory(const AvailableMetrics::Category& category)
 {
     bool category_has_data = false;
 
-    ImGui::BeginChild("scroll", ImVec2(-1, -1));
     for(const auto* table : category.ordered_tables)
     {
         uint64_t key = MetricId::GetTableKey(category.id, table->id);
@@ -280,8 +279,6 @@ ComputeTableView::RenderCategory(const AvailableMetrics::Category& category)
     {
         ImGui::TextDisabled("No data available for this category.");
     }
-
-    ImGui::EndChild();
 }
 
 ComputeTableView::Preset::Preset(ComputeTableView& widget)
--- a/src/view/src/compute/rocprofvis_compute_table_view.cpp
+++ b/src/view/src/compute/rocprofvis_compute_table_view.cpp
@@ -35,9 +35,7 @@ ComputeTableView::ComputeTableView(DataProvider&                     data_provid
     auto workload_changed_handler = [this](std::shared_ptr<RocEvent> e) {
         auto evt = std::dynamic_pointer_cast<ComputeSelectionChangedEvent>(e);
         if(evt && evt->GetSourceId() == m_data_provider.GetTraceFilePath())
-        {
             RebuildTabs();
-        }
     };
 
     m_workload_selection_changed_token = EventManager::GetInstance()->Subscribe(
@@ -91,6 +89,8 @@ ComputeTableView::~ComputeTableView()
 void
 ComputeTableView::RebuildTabs()
 {
+    if(m_tabs)
+        m_active_tab_id = m_tabs->GetActiveTab()->m_id;
     m_tabs.reset();
     m_table_widgets.clear();
     m_data_provider.ComputeModel().ClearKernelMetricValues(m_client_id);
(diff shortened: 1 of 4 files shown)


drchen-amd (dismissed):
- Tables should make use of available space and fill.Image
- Pinned table row headers can still be scrolled off screen.

drchen-amd (commented):
I still don't see this happening:

> The table header is now fixed when scrolling down.

Image

amokiche-amd:
> I still don't see this happening:
> 
> > The table header is now fixed when scrolling down.
> 
> Image

What about headers name freezing it the same behavor as in comparison table
I had a discussion with Tom about that and we decided to make it with the same behavor for consistency and add new story for future to add frosen header litle bit more complicated, that not in priority for now.
image
horisontal scroling agree i hadn't knew about this feature, fixed

tomk-amd:
Not sure if this has always been a bug, or it is new, but the bottom table (child window) can be scrolled off of the screen if horizonal gesture scrolling is used.  (two finger horizontal swipe on trackpad for example).

image

amokiche-amd:
> Not sure if this has always been a bug, or it is new, but the bottom table (child window) can be scrolled off of the screen if horizonal gesture scrolling is used. (two finger horizontal swipe on trackpad for example).
> 
> image

image
Yes, it's been like this from the very beginning. The reason is that comparison tables use the ImGuiTableFlags_ScrollX flag, while metric tables don't. However, if we use this flag, the table will no longer fill all the available space on the right, and we'll see a large empty space. This isn't a problem for comparison tables, since they have a large number of columns.
If we add this flag to metric tables, they will look like this:

drchen-amd:
> I had a discussion with Tom about that and we decided to make it with the same behavor for consistency and add new story for future to add frosen header litle bit more complicated, that not in priority for now.

In that case I will let Tom approve since I don't know the details of what was discussed. I think current state is fine. The pinned section is now readable, and the active tab persisted across workload switch. 

> What about headers name freezing it the same behavor as in comparison table

If you mean the pinned section, then yes with latest commits it is the same as comparison. 

For the other tables, they do not have the same behavior. Each table in comparison becomes scrollable only when there is not enough space, and the scrollable size is scaled to fit such that all row/column headers and scroll bars are visible. 

For this PR, each table becomes scrollable if they exceed 40 rows. This does not help you if you do not have enough space to display 40 rows because any amount of scrolling will push the row header off screen. It also may not be ideal if you have more than 40 rows of space, because then you're forcing a scrollbar when it is not be required.

drchen-amd:
> Not sure if this has always been a bug, or it is new, but the bottom table (child window) can be scrolled off of the screen if horizonal gesture scrolling is used. (two finger horizontal swipe on trackpad for example).

This is a generic issue with the tab container. It happens even to the project level if you open enough projects or squish the window:
Screenshot 2026-05-01 113111

src/view/src/widgets/rocprofvis_compute_widget.cpp
```diff
@@ -60,7 +60,7 @@ MetricTableBase::Render()
     if(ImGui::BeginTable(RocWidget::GenUniqueName(table_id).c_str(), num_columns,
                          m_table_flags, table_hegth))
     {
-        ImGui::TableSetupScrollFreeze(m_freezed_columns, m_freezed_rows);
+        ImGui::TableSetupScrollFreeze(3, 1);
```
tomk-amd: Please remove magic numbers (add consts for these)

tomk-amd (approved):
Please define constants instead of magic numbers, otherwise looks good.

drchen-amd:
This change became broken again after approval:

- Table view can only be scrolled by dragging scroll bar, it can not be scrolled using mouse wheel/trackpad. 
- Category tab bar can be scrolled off screen.

Screenshot from 2026-05-07 10-17-44

4b7c8a188ac98a1165dccebf29d2cb36d9f9ed6a - Fail
a3a8717ef4526611b17f1c951e15c93952663b08 - Pass

src/view/src/compute/rocprofvis_compute_table_view.cpp
```diff
@@ -91,6 +91,8 @@ ComputeTableView::~ComputeTableView()
 void
 ComputeTableView::RebuildTabs()
 {
+    if(m_tabs)
+        m_active_tab_id = m_tabs->GetActiveTab()->m_id;
     m_tabs.reset();
```
tomk-amd, line 96: GetActiveTab() can return null - this code crashes!

PR #818, opened 2026-05-05
Total App Redesign 

Sajeeth-Wimalasuriyan:
App-redesign branch focused on three areas:

- Visual redesign of the app shell — refreshed palette, tab/split chrome, timeline controls, compute surfaces, sticky notes, font stack, and styling defaults so the UI reads as one coherent product.

- Memory-chart rework — kernel-details memory chart redrawn with a grid backdrop, proper N/A handling, alpha-blended block accents, dashed flow lines, and direction-aware arrows that match the standard Omniperf flow (cache to CU for reads/requests, CU to cache/memory for writes/atomics).

- Stability / housekeeping — fixes the ImGui assertion crash that hit on every compute-comparison load, drops the buggy Windows DWM title-bar dark-mode hook, replaces the per-target CI workflows with per-OS files, and tightens render paths.

image
image


src/app/src/main.cpp
```diff
@@ -10,6 +10,7 @@
 #define GLFW_INCLUDE_NONE
 #include "AMD_LOGO.h"
 #include "rocprofvis_cli_parser.h"
+#include "rocprofvis_settings_manager.h"
 #include "rocprofvis_version.h"
```
tomk-amd: don't include settings manager.

tomk-amd (commented):
Image

First impressions:

The gap between the menu bar and the toolbar is too big / unnecessary.
I don't think the toolbar should be rounded? yellow highlight.
On the memory chart, I think the grid background is unnecessary and does not match the feel of the rest of the application.

tomk-amd:
There is a lot of extra white space at the bottom of compute summary page:

image

Same is true for kernel details, and table view.  I think the table widget might be adding some padding?

tomk-amd:
Track "name" is split across mulitple lines, causing it to clip:  
image

drchen-amd:
I think the memory chart looks very good, but otherwise there are lots of rough edges/inconsistencies with the details that lack polish. I won't attempt to list everything but  as a start, I consider these as downgrades vs dev...
- Toolbar has a different height in compute vs system. 
- Roof line menu toggle and options buttons are misaligned.
- Table row hover highlight is inconsistent. In compute tables its very faint vs an obvious blue in systems.
- Not a fan of compute summary chart having a backdrop. Makes the pie/bar buttons blend into the bg.
- Compute comparison pinned section and unpinned sections have different widths because the pinned section does not have the border the unpinned section does. Also the height is no longer a multiple of row height and causes the last row to either get cut off or leave a gap.
- Background tabs look greyed out/disabled.

drchen-amd:
> Stability / housekeeping — fixes the ImGui assertion crash that hit on every compute-comparison load

Is this on dev?

tomk-amd:
> > Stability / housekeeping — fixes the ImGui assertion crash that hit on every compute-comparison load
> 
> Is this on dev?

No I don't believe so... this seems to be a recurring artifact that appears when spacing / padding is adjusted in a window (child) without adjusting the amount a subsequent SetCursorPos adjusts the cursor position by.. an ImGUI assertion is hit. (cursor moves out of window bounds).

tomk-amd:
image
drop down color is the same as panel making controls invisible

drchen-amd:
I will polish the compute comparison. Unrelated to that area:

- Compute workload details has broken scrolling. The main tab bar can be scrolled out of view.
- Systems table filter text box "x" button displayes out of bounds. Should be like search bar.
- Systems annotation tab table is inconsistent with other tables. Row height seems larger and text is not vertically centered like the rest.
- Landing page open button bg and recent files hover highlighting is very faint.

drchen-amd (changes requested):
- Histogram context menu does not have standard padding.
- System summary node and gpu drop downs have very small heights vs other dropdowns.
- Event and sample tables get clipped prematurely as size is increased.

drchen-amd (changes requested):
- Systems table aggregate and group columns "x" buttons are displayed out of bounds. Should be like search bar.

drchen-amd (changes requested):
- Frame color of compute table checkboxes lacks contrast with even row bg. Unchecked checkboxes are pretty much invisible in dark mode.
- Compute table view row hover highlight only works if unit column is hovered.

drchen-amd (changes requested):
Annotation tab still seems off:
Image

- Small scroll bar in title column.
- Text supposed to go in text column?
- Last row is fat.

src/view/src/widgets/rocprofvis_tab_container.cpp
```diff
@@ -92,12 +92,34 @@ TabContainer::Update()
...
+    SettingsManager& settings = SettingsManager::GetInstance();
+    const ImGuiStyle& style = settings.GetDefaultStyle();
+
+    // Keep active tabs visually connected to the content panel.
+    ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding,
+                        ImVec2(style.WindowPadding.x + 2.0f, style.WindowPadding.y));
+    ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing,
+                        ImVec2(style.ItemSpacing.x * 0.6f, style.ItemSpacing.y * 0.5f));
+    ImGui::PushStyleVar(ImGuiStyleVar_FramePadding,
+                        ImVec2(style.FramePadding.x + 2.0f, style.FramePadding.y + 1.0f));
```
drchen-amd, line 104: I really don't like these unique offsets applied to the style vars because it will make keeping things in sync troublesome. For example, I had to randomly add 2.0f to the padding in comparison to make it match the styling used by the tabs.

Sajeeth-Wimalasuriyan:
> Annotation tab still seems off: Image
> 
> * Small scroll bar in title column.
> * Text supposed to go in text column?
> * Last row is fat.

yes Im addressing this rn seems like a regression

drchen-amd (dismissed):
Event table/sample table aggregate dropdown X button is now gone. Used to appear to the left of the dropdown arrow when something other than None was chosen.

tomk-amd (approved):

drchen-amd:
Since its already approved, I'll just mention some final observations here
- Annotation tab: title and text columns are top aligned; time and visible columns are center aligned. Should all be center aligned.
- Compute metric tables: row hover highlighting only works if unit column is hovered. 
- The new table row bg has poor contrast with item frames on even rows. 
  - Ex: Checkbox for pinning compute metrics is barely visible in light mode and is invisible in dark mode.
  - Ex: Settings panel hotkey table. Same as above.
-  Settings panel hotkey page has a non-functional double scrollbar above size 22 font.
- Table aggregate dropdown X button:
  -  Is a grey shade while others are solid black. 
  - Clicks through to the dropdown (clicking the X also triggers the dropdown)

tomk-amd:
> Table aggregate dropdown X button:
> * Is a grey shade while others are solid black.
> * Clicks through to the dropdown (clicking the X also triggers the dropdown)

This still happens though, I didn't notice it, it would be nice to fix it

drchen-amd:
Yes, every item I listed still happens.

PR #819, opened 2026-05-06
Use the new font uploading aproach for the fonts

amokiche-amd:
## Motivation
Use the modern imgui approach for loading fonts, for dynamic font resizing.

## Technical Details
Now only two fonts are loaded: the main font and the icon font. When using PushFont, you must specify the font size, or 0.0f if the font size should not be changed.

## Test Plan
Check that fonts work correctly on different sizes and systems, and that icons are displayed correctly.

## Submission Checklist

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -501,11 +501,11 @@ AppWindow::RenderEmptyState()
     }
 
     // --- Title ---
-    ImFont* title_font = settings.GetFontManager().GetFont(FontType::kLarge);
-    if(title_font) ImGui::PushFont(title_font, settings.GetFontManager().GetFontSize(FontType::kLarge));
+
+    ImGui::PushFont(NULL, settings.GetFontManager().GetFontSize(FontSize::kLarge));
     CenterNextTextItem("Open a trace or project");
     ImGui::TextUnformatted("Open a trace or project");
-    if(title_font) ImGui::PopFont();
+    ImGui::PopFont();
 
     ImGui::Dummy(ImVec2(0.0f, font_size * 0.25f));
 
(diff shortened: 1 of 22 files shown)


src/view/src/rocprofvis_font_manager.cpp
```diff
@@ -141,73 +118,64 @@ FontManager::Init()
...
 {
-    if(static_cast<int>(font_type) < 0 ||
-       static_cast<int>(font_type) >= static_cast<int>(FontType::__kLastFont))
-    {
-        return nullptr;  // Invalid font type
-    }
-    return m_fonts[static_cast<int>(font_type)];
+    (void)font_type;
+    return m_text_font;
 }
```
tomk-amd: This method, `GetFont()`, should be re-purposed.  FontType - as it is right now, is only used to control the size of the font.    

I'd recommend creating a new enum called FontSize, where kLarge, kSmall, etc.. are defined.  Then repurpose FontType to have items like, kSansSerif, kMonoSpace, and possibly kIcon..   kDefault for FontType would be kSansSerif.

In the previous implementation we needed FontType tied to the size as we were creating different 'fonts' for each size.

src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -502,7 +502,7 @@ AppWindow::RenderEmptyState()
 
     // --- Title ---
     ImFont* title_font = settings.GetFontManager().GetFont(FontType::kLarge);
-    if(title_font) ImGui::PushFont(title_font);
+    if(title_font) ImGui::PushFont(title_font, settings.GetFontManager().GetFontSize(FontType::kLarge));
     CenterNextTextItem("Open a trace or project");
```
tomk-amd: Related to comment about the FontType enum, I think all of these PushFont() calls become, PushFont(NULL,font_size), instead of PushFont(font_type, font_size);  The font_type only would get set if we want to change the font_face.

tomk-amd (approved):

drchen-amd:
Was this intentionally put on main and not on dev?

drchen-amd:
Nevermind I see theres a cherrypick in progress.

tomk-amd:
> Nevermind I see theres a cherrypick in progress.

Nope our mistake, hence the cherry pick

amokiche-amd:
My bad, I chose wrong branch, but cherry-pick to dev is ready

PR #823, opened 2026-05-08
Don't block UI thread during DataProvider teardown

tomk-amd:
## Motivation

Do not block the UI thread when tearing down the data provider.  Waiting for the backend to cancel / complete in-flight requests can take a long time, and this was done in a way that blocks the UI thread, causing the dreaded "Application not responding" message from the OS.

## Technical Details

Move DataProvider request/controller cleanup into detachable cleanup work object so tab close and app shutdown can hand long waits to AppWindow background jobs instead of blocking the UI thread.

Add shutdown cleanup polling, persistent progress notifications, and a centered modal shutdown message so the app stays responsive while outstanding requests are cancelled and freed.

Add loading-indicator centering options, including inline rendering support, to `RenderLoadingIndicator` and update the shutdown dialog to use the new inline/centered layout.

Added better feedback if Database cleanup option fails to start. (Previously no user feedback was present).

## Todo

Database cleanup will still cause UI thread to block as it causes the Dataprovider to abort requests using a different path.  This will be refactored in a separate PR.

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -53,6 +53,9 @@ constexpr const char* SUPPORTED_FILE_TYPES_HINT   = "Supported types: .db, .rpd,
 
 constexpr float STATUS_BAR_HEIGHT = 30.0f;
 
+constexpr const char* CLEANUP_MESSAGE = "Waiting for requests to finish cleanup...";
+constexpr const char* CLOSING_MESSAGE = "Closing...";
+
 // For testing DataProvider
 void
 RenderProviderTest(DataProvider& provider);
@@ -788,6 +791,9 @@ AppWindow::RenderShutdownState()
         ImGuiCond_Always, ImVec2(0.5f, 0.5f));
     ImGui::SetNextWindowSize(ImVec2(360.0f * dpi, 0.0f), ImGuiCond_Always);
 
+    PopUpStyle ps;
+    ps.PushPopupStyles();
+    ps.CenterPopup();
     if(ImGui::BeginPopupModal(SHUTDOWN_DIALOG_NAME, nullptr,
                               ImGuiWindowFlags_NoResize |
                                   ImGuiWindowFlags_NoMove |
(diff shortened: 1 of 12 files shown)


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -959,8 +1194,10 @@ AppWindow::HandleTabClosed(std::shared_ptr<RocEvent> e)
             }
         }
         spdlog::debug("Tab closed: {}", tab_closed_event->GetTabId());
-        m_projects[tab_closed_event->GetTabId()]->Close();
-        m_projects.erase(tab_closed_event->GetTabId());
+        project_it->second->Close();
+        DetachProjectProviderCleanup(*project_it->second,
+                                     ProviderCleanupReason::kTabClose);
```
drchen-amd, line 1214: Do you think the tab close case needs the notification? User closes a tab, and the tab disappears immediately. Does user need to know about the progress of this invisible bg task we're running when it does not affect their usage of the app? If not, this could be dev mode only like the compute metric fetch notifications.

For the app close scenario, it makes sense to notify, because it provides justification for why app stays around after being closed.

tomk-amd: Yeah I was debating the utility of the notifications.  For app close I decided that the notifcations were not obvious enough, and hence the "Shutdown" screen with the dialog showing that the app is waiting for requests.    

But re-visiting this I did realize that one of the notifications was not being displayed properly.. The general persistent closing trace notification... It was incorrect to "show" it every frame (opacity keeps getting reset to 0 as part of fade-in animation) making it invisible..

tomk-amd: In terms of whether to show the notification on tab close, I think the valid use case for it is the following:  

Open larger trace, close trace while it loading (this is the operation that currently takes the longest to tear down).  Wait x seconds or even minutes.  Close application, get message saying waiting for trace(s) to close.  From the user's perspective, "What trace?? I don't have any trace open?".

src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -371,19 +392,176 @@ AppWindow::GetCurrentProject()
...
+    }
+
     HotkeyManager::GetInstance().ProcessInput();
     EventManager::GetInstance()->DispatchEvents();
     DebugWindow::GetInstance()->ClearTransient();
     m_tab_container->Update();
 #ifdef ROCPROFVIS_DEVELOPER_MODE
     m_test_data_provider.Update();
 #endif
+    UpdateProviderCleanups();
```
drchen-amd: Is this second call needed?

drchen-amd (approved):

src/view/src/rocprofvis_appwindow.cpp
tomk-amd: Nope, good catch.

PR #826, opened 2026-05-11
Polish Redesigned UI

Sajeeth-Wimalasuriyan:
-Refined dark-mode event details colors to remove the overly bright white panel/background.
-Added separate theme colors for table borders, panel borders, and table row/header backgrounds.
-Fixed stale event-detail panel backgrounds by allowing split container items to inherit the active theme background.
-Aligned Annotation table row content so title/text/time/visibility cells are vertically consistent.
-Fixed compute metric table row hover so rows highlight across all cells, not just the Unit column.
-Improved control contrast on table rows so checkboxes/buttons remain visible in light and dark mode.
-Fixed Settings hotkey table sizing to avoid duplicate scrollbars at larger font sizes.
-Fixed Aggregate dropdown clear X styling and click behavior so clearing does not also open the dropdown.


src/view/src/rocprofvis_multi_track_table.cpp
```diff
@@ -177,45 +177,68 @@ MultiTrackTable::Render()
...
+        const ImVec2 combo_min     = ImGui::GetCursorScreenPos();
+        const ImVec2 combo_max(combo_min.x + ImGui::CalcItemWidth(),
+                               combo_min.y + ImGui::GetFrameHeight());
+        const ImVec2 clear_min(combo_max.x - combo_arrow_w - clear_button_w,
+                               combo_min.y);
+        const ImVec2 clear_max(combo_max.x - combo_arrow_w, combo_max.y);
+        const bool has_group_by_selection = (m_group_by_selection_index != 0);
+        const bool clear_region_hovered =
+            has_group_by_selection &&
+            ImGui::IsMouseHoveringRect(clear_min, clear_max, false);
```
tomk-amd: Creating a button effect around the clear icon looks nice, but now this is inconsistent with the clear icon's behaviour around the rest of the app, even in this same control set.  For example the Filter input also shows a clear icon when there is text present, but it does not highlight when hovered.

tomk-amd (approved):

PR #832, opened 2026-05-12
Enhance Flow Data table with hover and origin highlights

Sajeeth-Wimalasuriyan:
The Flow Data table previously had no visual feedback on hover, and there was no way to tell which row corresponded to the event you had actually clicked on to open the view. This PR brings the Flow Data table up to parity with the other tables in the app:

-Row hover highlight — Hovering any flow row paints the row background (same kHighlightChart color used elsewhere) and broadcasts a timeline highlight through the existing TimelineSelection pipeline, so the corresponding event lights up on the timeline as well.
-Origin (owner) row indicator — The row representing the event you opened the view for is tinted with a new kAreaOfInterest theme color, and hovering it shows a tooltip explaining what it is.
-New shared theme color — Colors::kAreaOfInterest (dark + light) added to the central palette so future "this is the focal element" tints stay consistent across the app.

image


src/view/src/rocprofvis_events_view.h
```diff
@@ -57,6 +59,14 @@ class EventsView : public RocWidget
...
     int                                      m_context_menu_flow_index;
     int                                      m_context_menu_flow_column;
+    // Flow row hovered last frame. Used to paint the row background
+    // (InfiniteScrollTable-style) and to drive the timeline event highlight.
+    uint64_t                                 m_flow_hover_owner_event_id;
+    uint64_t                                 m_flow_hover_flow_event_id;
+    uint64_t                                 m_flow_hover_flow_track_id;
+    uint64_t                                 m_frame_flow_hover_owner_event_id;
+    uint64_t                                 m_frame_flow_hover_flow_event_id;
+    uint64_t                                 m_frame_flow_hover_flow_track_id;
```
tomk-amd: I think it would be better to wrap these up in a struct like FlowHighlightState or something similar.

src/view/src/rocprofvis_events_view.cpp
```diff
@@ -339,6 +392,20 @@ EventsView::RenderEventFlowInfo(const EventInfo* event_data)
...
+                            m_frame_flow_hover_owner_event_id = this_owner_event_id;
+                            m_frame_flow_hover_flow_event_id  = flow.id.uuid;
+                            m_frame_flow_hover_flow_track_id  = flow.track_id;
+                            if(flow.id.uuid == this_owner_event_id)
+                            {
+                                // Explain the tint on the owner row so the
+                                // user knows why this row looks different.
+                                SetTooltipStyled(
+                                    "This is the event you opened to view "
+                                    "this flow.");
```
tomk-amd: Not sure tooltip is neccessary

tomk-amd (approved):

PR #834, opened 2026-05-13
Fix annotation scroll interactions

Sajeeth-Wimalasuriyan:
-Fixed annotation scrollbar interactions so mouse wheel/scrollbar input inside an expanded annotation no longer also drives timeline scrolling or zooming.
-Updated sticky notes to report when they are consuming timeline input, allowing the timeline to ignore those interactions while the annotation is hovered.
-Limited expanded annotation dragging to the header area so grabbing the note body scrollbar scrolls the annotation content instead of moving the annotation.

### DIFF
--- a/src/view/src/rocprofvis_stickynote.cpp
+++ b/src/view/src/rocprofvis_stickynote.cpp
@@ -103,7 +103,7 @@ StickyNote::SetTitle(std::string title)
 {
     m_title = title;
 }
-void
+bool
 StickyNote::Render(ImDrawList* draw_list, const ImVec2& window_position,
                    std::shared_ptr<TimePixelTransform> tpt)
 {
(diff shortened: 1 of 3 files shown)


src/view/src/rocprofvis_stickynote.cpp
```diff
@@ -398,8 +403,10 @@ StickyNote::HandleDrag(const ImVec2&                       window_position,
         {
             drag_pos = ImVec2(window_position.x + x, window_position.y + y);
         }
+        // Expanded notes drag by the header so the scrollable body can own
+        // scrollbar/body interactions.
         drag_w   = m_size.x;
-        drag_h   = m_size.y;
+        drag_h   = 36.0f;
         drag_max = ImVec2(drag_pos.x + drag_w, drag_pos.y + drag_h);
     }
```
tomk-amd: What does this magic number represent?

Sajeeth-Wimalasuriyan: It's the expanded note header height. I replaced the magic number with a named constant and use it directly now.

tomk-amd (approved):

PR #841, opened 2026-05-18
AMokichev/ Add new widget to show code and correspond it to the ISA

amokiche-amd:
## Motivation
[Optiq-Compute] Add new widget to show code and correspond it to the ISA
Add the prototype for PC sampling view to analyze the code and ISA

## Technical Details
image

A data model for PC sampling has been added, along with data loading from the database via a controller and populating the data model. This data model is then used to build the code for ISA View. To enable this, you must build the tool with the ROCPROFVIS_DEVELOPER_MODE flag.

The part of the code responsible for the stall is a stub; there is currently no clear picture of what it should look like.

## Submission Checklist


drchen-amd:
I haven't tested/looked deeply yet, but it looks like all data is loaded once on initial file open?

I think this use case warrants breaking up the data load into an initial discovery, followed by dynamically loading pieces driven by user input, like how the metrics how handled. I mention this because... 

1. I imagine the total volume of data for this use case may be large. Going by schema, each kernel may have multiple source files and multiple isa files, each of which is a big string list. Within each isa file, each line may have multiple stall records.
2. We probably will not need to display all available data simultaneously. I assume we will only show one source file and one isa file, while listing the others as available.

src/controller/src/compute/rocprofvis_controller_trace_compute.cpp
```diff
@@ -490,12 +490,300 @@ rocprofvis_result_t ComputeTrace::LoadRocpd(Future* future)
...
+                                                    { kRPVComputeColumnPcSamplingCodeObjectChecksum, std::nullopt },
+                                                },
+                                                {}
+                                            };
+                                            future->ResetProgress();
+                                            ExecuteQuery(db, m_dm_handle, object2wait, nullptr, kRPVComputeFetchKernelCodeObjects, m_query_arguments, m_query_output, [this, &kernel](const QueryDataStore& data_store){
+                                                rocprofvis_handle_t* pc_handle = nullptr;
+                                                if(kRocProfVisResultSuccess != kernel->GetObject(kRPVControllerKernelPcSamplingIndexed, 0, &pc_handle) || !pc_handle)
+                                                    return;
+                                                ((PcSampling*)pc_handle)->SetUInt64(kRPVControllerPCSamplingNumCodeObjects, 0, data_store.rows.size());
```
tomk-amd: Maybe use Reference class to convert handle to object instead of  direct casting, ex:

```
   RocProfVis::Controller::SystemTraceRef  trace(controller_handle);
   if(trace.IsValid() {
     trace->AsyncFetch(...)
   }
```

Create a typedef for the new pc sampling and kernel classes ex:

`typedef Reference<rocprofvis_controller_pc_sampling_t, PcSampling , kRPVControllerObjectTypePcSampling> PcSamplingRef;`

src/controller/src/compute/rocprofvis_controller_kernel.cpp
```diff
@@ -88,6 +88,30 @@ rocprofvis_result_t Kernel::GetUInt64(rocprofvis_property_t property, uint64_t i
...
 
+rocprofvis_result_t Kernel::GetObject(rocprofvis_property_t property, uint64_t index, rocprofvis_handle_t** value)
+{
+    (void)index;
+    rocprofvis_result_t result = kRocProfVisResultInvalidArgument;
+    if(value)
+    {
+        switch(property)
+        {
+            case kRPVControllerKernelPcSamplingIndexed:
```
vstempen: It seems indexed property is used to get static data. A bit misleading.

amokiche-amd:
> I haven't tested/looked deeply yet, but it looks like all data is loaded once on initial file open?
> 
> I think this use case warrants breaking up the data load into an initial discovery, followed by dynamically loading pieces driven by user input, like how the metrics how handled. I mention this because...
> 
> 1. I imagine the total volume of data for this use case may be large. Going by schema, each kernel may have multiple source files and multiple isa files, each of which is a big string list. Within each isa file, each line may have multiple stall records.
> 2. We probably will not need to display all available data simultaneously. I assume we will only show one source file and one isa file, while listing the others as available.

Now I think I should have done this in the next pull request, but since I'd already started, I finished it right away. So this functionality has been added.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -4349,6 +4354,108 @@ void DataProvider::SetFetchMetricsCallback(
...
+            it = m_requests.erase(it);
+        }
+        else
+        {
+            ++it;
+        }
+    }
+
+    rocprofvis_controller_future_t*    future = rocprofvis_controller_future_alloc();
+    rocprofvis_controller_arguments_t* args   = rocprofvis_controller_arguments_alloc();
```
tomk-amd: Cancelling should not be done here I think...

Doing this:  `it = m_requests.erase(it);` will cause a mem leak as nothing cleans up the future / args from the request cancelled in this block..  ProcessPcSamplingRequest() will not be called!

Let the caller Cancel the request using DataProvider::CancelRequest() is busy..

Inside this function the logic should just be:
```
    auto it = m_requests.find(request_id );
    if (it == m_requests.end())
    { ... fetch ... }
    else return false;
```

The caller can do...

```
        if(m_data_provider.IsRequestPending(request_id))
        {
            m_data_provider.CancelRequest(request_id);
        }
        else
        {
        // request once clear ...
          m_data_provider.FetchPcSampling(..)
        }
```

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -4349,6 +4354,108 @@ void DataProvider::SetFetchMetricsCallback(
...
+            uint64_t workload_id = 0;
+            rocprofvis_controller_get_uint64(workload_handle, kRPVControllerWorkloadId, 0, &workload_id);
+            if(static_cast<uint32_t>(workload_id) != params.m_workload_id) continue;
+
+            uint64_t num_kernels = 0;
+            rocprofvis_controller_get_uint64(workload_handle, kRPVControllerWorkloadNumKernels, 0, &num_kernels);
+            for(uint64_t ki = 0; ki < num_kernels; ki++)
+            {
+                rocprofvis_handle_t* kernel_handle = nullptr;
+                rocprofvis_controller_get_object(workload_handle, kRPVControllerWorkloadKernelIndexed, ki, &kernel_handle);
```
tomk-amd: This seems overly complicated, having to walk the tree each time to find the kernel handle..

maybe add convenience functions to the controller ex:

```
rocprofvis_handle_t* workload_handle = nullptr;
rocprofvis_handle_t* kernel_handle   = nullptr;
rocprofvis_handle_t* pc_handle       = nullptr;

// 1. workload by id (off the trace controller root)
rocprofvis_controller_get_object(
    m_trace_controller, kRPVControllerWorkloadById, workload_id, &workload_handle);

// 2. kernel by id (off the workload)
if(workload_handle)
    rocprofvis_controller_get_object(
        workload_handle, kRPVControllerWorkloadKernelById, kernel_id, &kernel_handle);

// 3. pc-sampling handle (already a direct accessor)
if(kernel_handle)
    rocprofvis_controller_get_object(
        kernel_handle, kRPVControllerKernelPcSampling, 0, &pc_handle);

if(!pc_handle) return false;
```

kRPVControllerWorkloadById, kRPVControllerWorkloadKernelById accessors need to be added controller side.

This doesn't change anything in terms of performance still have to iterate through the list but the mess moves controller side, ex:

```
case kRPVControllerWorkloadById:
{
    result = kRocProfVisResultOutOfRange;
    for(auto* workload : m_workloads)
    {
        uint64_t id = 0;
        if(workload->GetUInt64(kRPVControllerWorkloadId, 0, &id) == kRocProfVisResultSuccess && id == index)
        {
            *value = (rocprofvis_handle_t*)workload;
            result = kRocProfVisResultSuccess;
            break;
        }
    }
    break;
}
```

Perhaps on the controller instead of using vectors, we use a map then perf would improve, but not sure what the side effects would be at the moment.

src/controller/src/compute/rocprofvis_controller_trace_compute.cpp
```diff
@@ -309,6 +313,256 @@ rocprofvis_result_t ComputeTrace::AsyncFetch(Table& table, Arguments& args, Futu
...
+    }
+
+    future.Set(JobSystem::Get().IssueJob([this, &output, kernel_id, source_file_id](Future* future) -> rocprofvis_result_t {
+        rocprofvis_dm_database_t db = rocprofvis_dm_get_property_as_handle(m_dm_handle, kRPVDMDatabaseHandle, 0);
+
+        FetchCodeObjectsAndIsaLines(db, future, kernel_id, output);
+        FetchIsaLineDepsAndStalls(db, future, output);
+        FetchSourceFileAndLines(db, future, source_file_id, output);
+
+        if(future->IsCancelled())
```
tomk-amd, line 358: This cancel handling doesn't do much, as the all of the hard (long) work is already done by this point..  Mentioning this because of the issues of cancelling on the ui (dataprovider) side.   When one request is cancelled by the UI it remains in flight for a long time on the controller..  If a new one is started before the "cancelled" actually terminates, then both operations could potentially write to the same kernel's `m_pc_sampling_data ` member.

src/controller/src/compute/rocprofvis_controller_trace_compute.cpp
```diff
@@ -309,6 +313,256 @@ rocprofvis_result_t ComputeTrace::AsyncFetch(Table& table, Arguments& args, Futu
...
+                        if(output.QueryToPropertyEnum(column.first, property, type))
+                            SetObjectProperty((rocprofvis_handle_t*)&output, property, isa_offset + i, data_store.rows[i][column.second.value()], type);
+                isa_offset += data_store.rows.size();
+            });
+    }
+}
+
+void
+ComputeTrace::FetchIsaLineDepsAndStalls(rocprofvis_dm_database_t db, Future* future, PcSampling& output)
+{
```
tomk-amd, line 425: These functions are generally inefficient, executing mulitple queries per ISA line for example...

To fix properly though, I think the queries themselves need to be reworked.  Ex: ISA→ISA deps for the whole kernel..

```SQL
SELECT j.dependent_isa_line_id, j.dependency_isa_line_id
FROM isa_line_to_isa_line_junction j
JOIN isa_lines il ON il.id = j.dependent_isa_line_id
JOIN code_objects co ON co.id = il.code_object_id
WHERE co.kernel_id = <kernel_id>;
```

We can work with Vlad (@vstempen) to improve the queries

src/view/src/compute/rocprofvis_compute_code_view.h
```diff
@@ -0,0 +1,154 @@
...
+    ~BaseCodeWidget() = default;
+
+    virtual void Render() override = 0;
+
+    bool IsStallShown() { return m_show_stall; };
+    void ChangeStallVisibility(bool show) { m_show_stall = show; };
+
+protected:
+    void CalcutlateLineNumberWidth(uint32_t count);
+    void PushStyles();
```
tomk-amd: Typo, should be CalculateLineNumberWidth

src/view/src/compute/rocprofvis_compute_code_view.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+    ImGui::BeginChild("ControlPanel", ImVec2(0.0f, topHeight), true);
+
+    ImVec2 start = ImGui::GetCursorPos();
+
+    ImGui::BeginGroup();
+
+    RenderSourceFileDropdown();
+
+    const float button_isa_width      = ImGui::CalcTextSize("Hide ISA").x      + ImGui::GetStyle().FramePadding.x * 2.0f; //TODO: thing how can avoid dublication of strings
```
tomk-amd: typo: dublication -> duplication  thing -> think

make these const defintions:  "Hide ISA", "Show ISA", etc..

src/view/src/compute/rocprofvis_compute_code_view.h
```diff
@@ -0,0 +1,156 @@
...
+    std::shared_ptr<SourceCodeWidget> m_source_code;
+    std::shared_ptr<IsaCodeWidget>    m_isa_code;
+    LayoutItem::Ptr                   m_isa_layout_item;
+    std::shared_ptr<HSplitContainer>  m_horizontal_split_container;
+
+    uint32_t                        m_current_source_file_id = 0;
+    uint32_t                        m_current_code_object_id = 0;
+    uint32_t                        m_current_kernel_id      = 0;
+    uint32_t                        m_current_workload_id    = 0;
+
```
tomk-amd, line 60: initialize in constructor (follow pattern in other classes)

Not sure about other ids, but m_current_workload_id and m_current_kernel_id should be set to ComputeSelection::INVALID_SELECTION_ID?

src/view/src/compute/rocprofvis_compute_code_view.cpp
```diff
@@ -0,0 +1,610 @@
...
+                                button_comments_width +
+                                ImGui::GetStyle().ItemSpacing.x * 2.0f;
+
+    ImGui::SameLine(ImGui::GetContentRegionAvail().x + ImGui::GetCursorPosX() - buttons_width);
+
+    if(ImGui::Button(m_isa_layout_item->m_visible ? hide_isa_str : show_isa_str))
+        m_isa_layout_item->m_visible = !m_isa_layout_item->m_visible;
+
+    ImGui::SameLine();
+    static bool show_metadata_enabled = false;
```
tomk-amd: should this be a member var instead? I don't think it should be shared across all instances (multiple compute tabs open)

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -4349,6 +4354,68 @@ void DataProvider::SetFetchMetricsCallback(
...
+    {
+        spdlog::debug("PC sampling request already pending for kernel {} file {}",
+                      params.m_kernel_id, params.m_source_file_id);
+        return false;
+    }
+
+    rocprofvis_controller_future_t*    future    = rocprofvis_controller_future_alloc();
+    rocprofvis_controller_arguments_t* args      = rocprofvis_controller_arguments_alloc();
+    rocprofvis_handle_t*               pc_handle = rocprofvis_controller_pc_sampling_alloc();
+
```
tomk-amd: Is this necessary, allocating pc sampling via the UI, ex: `rocprofvis_controller_pc_sampling_alloc()` ?

This has the side effect that if the request is in flight, and the tab is closed, the general request clean up manager will free the request, and any associated array, but the general cleanup function does not know about objects that are allocated.   In general the objects sitting in the request struct are "borrowed", the controller owns their life time.  In this case the UI is responsible.   

Either the request cleanup function will need to handle this special case, or better yet - don't allocate at all, and access the pc sampling object that is already part of the kernel object in the controller?

src/view/src/compute/rocprofvis_compute_code_view.cpp
```diff
@@ -0,0 +1,632 @@
...
+        m_show_metadata_enabled = !m_show_metadata_enabled;
+        m_source_code->ChangeStallVisibility(m_show_metadata_enabled);
+        m_isa_code->ChangeStallVisibility(m_show_metadata_enabled);
+    }
+
+    if(m_isa_layout_item->m_visible)
+    {
+        ImGui::SameLine();
+        static bool show_comments_enabled = false;
+        if(ImGui::Button(show_comments_enabled ? "Hide Comments" : "Show Comments"))
```
tomk-amd, line 246: This should be a memeber, otherwise this will enable / disable comments for all loaded files.

tomk-amd (approved):
I think there is a possibility of a race condition on the controller side when PC sampling data is fetched / written.   But this feature will be disabled for a while still, so there will be time to fix and refactor as needed. 

Let's merge for now and keep improving.

PR #850, opened 2026-05-26
Remove custom DPI handling, switch to ImGui auto-DPI font sizing

Sajeeth-Wimalasuriyan:
**Summary**
Hands off DPI handling to ImGui and tightens the user-facing font controls.

DPI: removed custom path, switched to ImGui auto-DPI. Dropped rocprofvis_view_set_dpi and the GLFW content-scale callback. The view now relies on ImGui's ConfigDpiScaleFonts / ConfigDpiScaleViewports for per-monitor scaling. Also removed dpi_based_scaling from DisplaySettings and its JSON key — SettingsManager no longer tracks DPI at all.
Multi-viewport wiring in main.cpp. Enables ImGuiConfigFlags_ViewportsEnable and the DPI flags, sets FontScaleMain = 1.0f, and renders platform windows via UpdatePlatformWindows / RenderPlatformWindowsDefault (preserves the GLFW context for OpenGL).
Font setting: dropdown → clamped slider. Replaced the font dropdown with a 13-18 pt slider. User settings only persist a base font index. FontManager now exposes GetMinUserFontSize / GetMaxUserFontSize / GetFontSizeAt / GetDefaultFontSizeIndex / ClampFontSizeIndex so the panel never indexes into the raw size table.
Fixed FontSize indexing bug. ImGui's default font now uses FontSize::kDefault (the user-chosen base) instead of FontType::kDefault, which had been silently mapping to FontSize::kSmall (base − 1). Added static_asserts to keep kSizeOffsets in lockstep with the FontSize enum.

**Test plan**

 Windows debug build runs and the app launches with no console errors.

 Open Settings → Display: font slider clamps to 13-18 pt; preview matches the slider value.

 OK applies the size; Cancel reverts to the previous size.

 Reset returns the font to 15 pt.

 Settings JSON on disk loses dpi_based_scaling; existing saved configs still load (extra keys ignored).

 Drag the main window between monitors with different DPI: layout/fonts rescale via ImGui auto-DPI, no manual handling needed.

### DIFF
--- a/src/app/src/main.cpp
+++ b/src/app/src/main.cpp
@@ -234,6 +234,9 @@ main(int argc, char** argv)
         // Create initial window with Vulkan hint (GLFW_NO_API) by default
         // The backend setup will recreate the window if OpenGL is needed
         glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
+#if defined(GLFW_SCALE_TO_MONITOR)  // GLFW 3.3+
+        glfwWindowHint(GLFW_SCALE_TO_MONITOR, GLFW_TRUE);
+#endif
         GLFWwindow* window = glfwCreateWindow(RocProfVis::View::DEFAULT_WINDOWED_WIDTH,
                                               RocProfVis::View::DEFAULT_WINDOWED_HEIGHT,
                                               APP_NAME, nullptr, nullptr);
--- a/src/app/src/main.cpp
+++ b/src/app/src/main.cpp
@@ -43,15 +43,6 @@ drop_callback(GLFWwindow* window, int count, const char* paths[])
 }
 
 static void
-content_scale_callback(GLFWwindow* window, float xscale, float yscale)
-{
-    // Unused parameters
-    (void) window;
-    (void) yscale;
-    rocprofvis_view_set_dpi(xscale);
-}
-
-static void
 close_callback(GLFWwindow* window)
 {
     g_render_options =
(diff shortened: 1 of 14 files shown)


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -162,6 +162,7 @@ AppWindow::Init()
     LayoutItem status_bar_item(-1, initial_status_bar_height);
     status_bar_item.m_item =
         std::make_shared<RocCustomWidget>([this]() { RenderStatusBar(); });
+    status_bar_item.m_window_flags = ImGuiWindowFlags_NoScrollbar;
     LayoutItem main_area_item(-1, -initial_status_bar_height);
```
tomk-amd, line 165: Hmm this feels like a bandaid solution.. the status bar content should not overflow causing the scrollbar to appear in the first place..

HandleFontChanged() should re-calculate the height of the status area to the correct size.  I think that maybe this calculation is not correct after the DPI changes?

tomk-amd:
Something strange is happening with this new auto DPI scaling flags:
                io.ConfigDpiScaleFonts             = true;
                io.ConfigDpiScaleViewports         = true;

When you adjust the window size, the font size changes subtly sometimes..  This is probably the cause of the status bar needing a scrollbar.  It's size is not recalculated because there is no notification that the font size has changed.

*Edit* - hmm not sure about this anymore, it might be an illusion, as the window texture is frozen and scaled during the drag operation, so the font distorts during the drag..

src/app/src/main.cpp
```diff
@@ -243,9 +234,6 @@ main(int argc, char** argv)
         // Create initial window with Vulkan hint (GLFW_NO_API) by default
         // The backend setup will recreate the window if OpenGL is needed
         glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
-#if defined(GLFW_SCALE_TO_MONITOR)  // GLFW 3.3+
-        glfwWindowHint(GLFW_SCALE_TO_MONITOR, GLFW_TRUE);
-#endif
```
tomk-amd, line 248: Put this back please

PR #853, opened 2026-05-27
Improve splitter and tab visibility, rename misleading accent enum

dhingora-amd:
## Motivation
The horizontal splitter between the timeline and the tab panel was effectively invisible, so users couldn't tell it was draggable. The active tab in the bottom tab bar also blended into the surrounding panel, making the selection hard to identify. Also, the `Colors::kAccentRed*` enum was misleading since the color is actually blue.

## Technical Details
- Splitter handles now paint a subtle bar at rest and highlight with `kAccent` on hover/drag, applied to both `SplitContainerBase::Render()` and the sidebar splitter in `TimelineView::RenderSplitter()`.
- Slightly darkened `kSplitterColor` in both themes.
- Switched unselected tab backgrounds in `TabContainer::Render()` from `kBgFrame` to `kButton` so inactive tabs are distinct from the panel.
- Renamed `Colors::kAccentRed{,Hover,Active}` → `Colors::kAccent{,Hover,Active}`. No color values changed.


src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -712,22 +712,33 @@ TimelineView::RenderSplitter()
 
     ImVec2 display_size = ImGui::GetWindowSize();
 
-    ImGui::SetNextWindowSize(ImVec2(1.0f, display_size.y), ImGuiCond_Always);
+    const float splitter_width = 5.0f;
```
drchen-amd: I would move this to the top of the file where the other constants are.

drchen-amd (approved):

Sajeeth-Wimalasuriyan (approved):
Hey, reviewed earlier but forgot to drop a approve. It's all good and nice job on the naming fix.

PR #857, opened 2026-05-29
Add copy menu to Call Stack table and fix cell right-click hitboxes

dhingora-amd:
## Motivation
The Flow table in the Event Details tab has a right-click "Copy Row Data" / "Copy Cell Data" menu, but the Call Stack table did not. Also, in both tables, Copy Cell would often grab the wrong column when right-clicking inside the cell — making the feature unreliable.
## Technical Details
- Added "Copy Row Data" and "Copy Cell Data" entries to the Call Stack table's right-click context menu, mirroring the existing Flow table behavior. Tracks the clicked column via a new `m_context_menu_callstack_column` member.
- Fixed the cell hitbox bug in both tables: cell text only covers as much width as the string, so right-clicks on the surrounding padding fell through to the row-spanning `Selectable` and were hardcoded to column 0. Now uses `ImGui::TableGetHoveredColumn()` to record the actual column under the cursor.

### DIFF
--- a/src/view/src/rocprofvis_events_view.cpp
+++ b/src/view/src/rocprofvis_events_view.cpp
@@ -21,6 +21,9 @@ constexpr ImGuiTableFlags TABLE_FLAGS = ImGuiTableFlags_BordersOuter |
                                         ImGuiTableFlags_Resizable |
                                         ImGuiTableFlags_SizingStretchProp;
 
+constexpr int kFlowColumnCount      = 6;
+constexpr int kCallStackColumnCount = 5;
+
 namespace
 {
 
@@ -327,7 +330,7 @@ EventsView::RenderEventFlowInfo(const EventInfo* event_data)
         }
         else
         {
-            if(ImGui::BeginTable("FlowInfoTable", 6, TABLE_FLAGS))
+            if(ImGui::BeginTable("FlowInfoTable", kFlowColumnCount, TABLE_FLAGS))
             {
                 ImGui::TableSetupColumn("ID");
                 ImGui::TableSetupColumn("Name");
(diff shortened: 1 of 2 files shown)


src/view/src/rocprofvis_events_view.cpp
```diff
@@ -402,8 +403,10 @@ EventsView::RenderEventFlowInfo(const EventInfo* event_data)
                                 ImGuiHoveredFlags_AllowWhenOverlappedByItem);
                         if(ImGui::IsItemClicked(ImGuiMouseButton_Right))
                         {
-                            m_context_menu_flow_index  = i;
-                            m_context_menu_flow_column = 0;
+                            m_context_menu_flow_index = i;
+                            int hovered_col           = ImGui::TableGetHoveredColumn();
+                            m_context_menu_flow_column =
+                                (hovered_col >= 0 && hovered_col < 6) ? hovered_col : 0;
                             ImGui::OpenPopup("##FlowContextMenu");
```
tomk-amd: Instead of magic numbers, please define a constant like kFlowColumnCount = 6 and use it instead.

src/view/src/rocprofvis_events_view.cpp
```diff
@@ -671,6 +678,28 @@ EventsView::RenderCallStackData(const EventInfo* event_data)
...
+                        std::string cells[] = {
+                            std::to_string(ctx_frame.id.bitfield.event_id),
+                            ctx_frame.address, ctx_frame.name, ctx_frame.file,
+                            ctx_frame.pc};
+
+                        if(ImGui::MenuItem("Copy Row Data"))
+                        {
+                            std::string row_text;
+                            for(int c = 0; c < 5; c++)
+                            {
```
tomk-amd, line 698: Same comment regarding magic numbers.. please define constant.

tomk-amd (changes requested):
Looks good, except for the magic numbers

tomk-amd (approved):

PR #858, opened 2026-06-02
Add right-click copy for timeline event names and details

Sajeeth-Wimalasuriyan:
image

### DIFF
--- a/src/view/src/rocprofvis_click_manager.cpp
+++ b/src/view/src/rocprofvis_click_manager.cpp
@@ -18,7 +18,6 @@ TimelineFocusManager::GetInstance()
 TimelineFocusManager::TimelineFocusManager()
 : m_layer_focused(Layer::kNone)
 , m_all_layers_focused()
-, m_right_click_layer(Layer::kNone)
 {}
 
 void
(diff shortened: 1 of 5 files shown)


tomk-amd:
The behaviour is a bit awkward.  (This is due to our kludgy mouse z-index detection)

When an event is hovered and the user right clicks, the "copy event name" and "copy event details" options are not shown in the context menu.  (Because this activates only when an event is selected).

Once an event is selected and the user hovers over the event and right clicks, the copy menu options are shown. So far so good.

If the user hovers over a completely different event and right clicks, the copy menu options are still shown. Sort of confusing because when clicked it will copy the selected event's name or details not the one currently hovered.

If the user hovers over an empty area and right clicks, the event copy menu is gone.  This would make sense if the hover right click behaviour was tied to the an actual hovered event, but it is tied to the selected event(s).  Since the mouse doesn't need to be hovered over the event in question to perform the action the interaction feels strange.  

Why can I copy when hovered over some other arbitrary event, even on a different track, but not when hovered over empty space?   The only real qualifier for this portion of the context menu becoming available should be whether an event is selected or not. (Unless we improve the z-index mouse detection). The same goes for the "mark range" option.

Perhaps the timeline context menu should look like this:

With Selected Event  <-- heading, make text light gray or add different bg color to indicate that it's a heading
Copy Name
Copy Details
Mark Range
----------------------  <-- the portion above this separator only shown if an event is selected
Add Annotation
Enter Measurement Mode

tomk-amd (approved):

PR #862, opened 2026-06-03
Lazy Render on Idle — near-zero CPU/GPU at rest, instant on input

Sajeeth-Wimalasuriyan:
### TL;DR
We were redrawing the **entire** UI every single vsync interval — ~60–120 FPS — **forever**, even when the app was sitting completely idle. This PR makes rendering event-driven: when there's nothing to do, the app sleeps. When something happens, it wakes instantly.

**Idle render workload drops from ~60–120 FPS to 0 FPS — a ~100% cut in idle render cost**, with zero loss in responsiveness.

### Before vs after
| | Idle behavior | Idle frames/hour @60Hz |
|---|---|---|
| **Before** | `glfwPollEvents()` busy loop → full ImGui re-traversal + draw + GPU present every refresh (vsync-locked) | **~216,000** |
| **After** | `glfwWaitEvents()` parks the thread until a real OS event | **~0** |

A profiler is idle for most of its lifetime (reading a trace, thinking, sitting in the background). That entire window used to peg a CPU core and keep the GPU busy for nothing. Now it's free → cooler machine, quieter fans, longer battery, and dramatically less pixel traffic over SSH/remote-display sessions.

### How it works
- Idle → `glfwWaitEvents()` (sleeps at ~0% CPU). Busy → `glfwWaitEventsTimeout(1/60s)` so per-frame work keeps ticking.
- A new `WantsContinuousRender()` keeps the loop awake whenever there's real work: pending DataProvider requests, a provider in `kLoading`, queued (deferred-dispatch) events, cleanup jobs, active notifications, or the timeline's loading-timer debounce.
- A short post-input frame budget (4 frames) lets animations, the 1–2 frame deferred event dispatch, and ImGui's multi-frame layout settle before sleeping — so the final frame is never stale.

### Why it's safe (no "frozen UI" footguns)
Completed controller futures are detected by **polling once per frame**, and the event bus is **deferred-dispatch** — so a naive `glfwWaitEvents()` would have frozen trace loads and spinners. This PR explicitly keeps rendering through:
- the **entire** initial trace + topology load (`kLoading`, even when the pending-request count momentarily hits zero between stages),
- the timeline debounce window that gates track-data requests,
- in-flight async fetches, notifications, and shutdown/cleanup.

Net result: identical behavior under load and interaction, near-zero cost at rest.


src/app/src/main.cpp
```diff
@@ -322,7 +328,24 @@ main(int argc, char** argv)
...
+                    {
+                        g_frames_to_render = RENDER_FRAMES_AFTER_INPUT;
+                    }
+
+                    if(g_frames_to_render > 0)
+                    {
+                        // Busy: poll so per-frame controller/event work keeps running.
+                        glfwWaitEventsTimeout(BUSY_POLL_INTERVAL_SEC);
+                    }
+                    else
```
tomk-amd, line 343: I think we should just use `glfwPollEvents();` in this case.  No need for `glfwWaitEventsTimeout` and the interval `BUSY_POLL_INTERVAL_SEC` .. the application is in "active" state anyway.

src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -576,6 +576,58 @@ AppWindow::Update()
     UpdateStatusBar();
 }
 
+bool
+AppWindow::WantsContinuousRender()
+{
+    if(!m_provider_cleanup_jobs.empty() || m_disable_app_interaction ||
```
tomk-amd, line 582: I am a little bit unsure about this approach.  This function is going to keep growing more and more complex as new cases emerge that will need the application to be active. 

I wonder if some kind of opt-in system would be better?  

like:

```
void AppWindow::RequestActive() {
  s_active_count++;
}

void AppWindow::ReleaseActive() {
 s_active_count--;
}
```

Then 

```
bool AppWindow::WantsContinuousRender() {
  return s_active_count > 0;
}
```
I think we can leave as is for now, but something to think about...

src/app/src/main.cpp
Sajeeth-Wimalasuriyan: done

src/view/src/rocprofvis_appwindow.cpp
Sajeeth-Wimalasuriyan: will keep in mind

tomk-amd (approved):

PR #864, opened 2026-06-05
Add hover tooltip and copy menu to track meta area, fix per-track table count off-by-one

dhingora-amd:
## Motivation

The track meta area on every timeline track had no quick way to inspect a track's full identity or copy its name / ID. This PR adds a hover tooltip and a right-click "Copy" menu.

While verifying the tooltip's event/sample count against the per-track table, the table consistently showed one fewer row than the tooltip when an event's start timestamp landed exactly on the trace's upper bound. This PR fixes that too.

## Technical Details

- `TrackItem` builds a tooltip with Track ID, optional Node/PID, and the Events/Samples count (compact form for large counts), and shows it on hover of the empty meta area. A right-click popup adds "Copy track name / Copy track ID" with the standard copy-notification toast.
- Hover predicate uses the repo's standard `IsItemHovered + IsAnyItemHovered` idiom plus one bool the `Pill` already tracks, so gear-icon and pill tooltips still win where they should.
- `ProfileDatabase::BuildTableQuery` uses `<=` instead of `<` on the upper bound of the final chunk in its chunked time filter. Internal seams keep strict `<` so no row is double-counted. Per-track table counts now match the tooltip total.

### DIFF
--- a/src/view/src/rocprofvis_track_item.cpp
+++ b/src/view/src/rocprofvis_track_item.cpp
@@ -8,6 +8,8 @@
 #include "rocprofvis_utils.h"
 #include "spdlog/spdlog.h"
 #include "widgets/rocprofvis_gui_helpers.h"
+#include "widgets/rocprofvis_notification_manager.h"
+#include "widgets/rocprofvis_widget.h"
 #include <memory>
 #include <algorithm>
 
@@ -18,6 +20,8 @@ namespace View
 
 constexpr float kMinTooltipWrapWidth = 200.0f;
 constexpr float kMaxTooltipWrapWidth = 600.0f;
+
+inline constexpr uint64_t kCompactCountThreshold = 1000;
 
 float TrackItem::s_metadata_width = 400.0f;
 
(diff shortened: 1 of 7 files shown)


src/view/src/rocprofvis_track_item.cpp
```diff
@@ -369,6 +353,45 @@ TrackItem::RenderMetaArea()
...
+    const bool meta_area_hovered = ImGui::IsItemHovered() &&
+                                   !ImGui::IsAnyItemHovered() &&
+                                   !m_pill.WasLastHovered();
+
+    if(meta_area_hovered && !m_meta_area_tooltip.empty())
+    {
+        ImGui::SetNextWindowSizeConstraints(
+            ImVec2(0, 0), ImVec2(META_TOOLTIP_MAX_WIDTH, FLT_MAX));
+        BeginTooltipStyled();
+        ImGui::PushTextWrapPos(ImGui::GetCursorPosX() + META_TOOLTIP_MAX_WIDTH);
```
drchen-amd: This can be cut off: 

Image

Probably needs width - padding.

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -369,6 +353,45 @@ TrackItem::RenderMetaArea()
...
+        ImGui::TextUnformatted(m_meta_area_tooltip.c_str());
+        ImGui::PopTextWrapPos();
+        EndTooltipStyled();
+    }
+
+    const std::string copy_popup_id =
+        "##track_copy_menu_" + std::to_string(m_track_id);
```
drchen-amd: This is probably over kill to get an id since only one of these can be visible at the same time. Should be enough to define a constant like InfiniteScrollTable's ROWCONTEXTMENU_POPUP_NAME.

Also, the padding of this is not consistent vs others, its more cramped:

Image

Image

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -702,6 +725,38 @@ TrackItem::SetMetaAreaLabel(const TrackInfo* track_info)
...
+    {
+        meta_lines += "Node ID: " + node_id_str + "\n";
+    }
+    if(show_process_id)
+    {
+        meta_lines += "Process ID: " + process_id_str + "\n";
+    }
+    meta_lines += std::string(count_label) + ": " +
+                  std::to_string(track_info->num_entries);
+    if(track_info->num_entries >= META_TOOLTIP_COMPACT_COUNT_MIN)
```
drchen-amd, line 753: Maybe we don't need to display both full and compact and instead show only compact past your threshold?

"1906570 (1.91M)" 

Having the 1.91M improves readability, but I don't think it is useful to know 1906570 = 1.91M

drchen-amd (commented):
I'm noticing that some tracks have mismatched counts in the tooltips. 
Ex:
Screenshot from 2026-06-09 11-24-39
Screenshot from 2026-06-09 11-23-37

The cases I've found are all streams. 

Your change looks fine so its probably the count being off in the track info. The entry count field seems to be unused up until now.

(Also as I was testing this, maybe the full non compact count should always be shown in dev mode.)

tomk-amd:
> The cases I've found are all streams.
> 
> Your change looks fine so its probably the count being off in the track info. The entry count field seems to be unused up until now.
> 

I think it is ok to merge this with the stream count bug.  I think that should be addressed via a separate PR.

> (Also as I was testing this, maybe the full non compact count should always be shown in dev mode.)

Yes showing the full count in Dev mode would be useful

drchen-amd (approved):
Looks good. Please get @vstempen's approval as well especially for the last commit.

PR #865, opened 2026-06-05
Add support for LDS AI roofline plot point

jamessiddeley-amd:
## Motivation

Add support in roc-optiq for LDS roofline data produced by rocprofiler-compute schema `1.4.0` so Optiq can include LDS AI points in compute roofline views.

## Technical Details

Similar to the paths of AI L1, L2, HBM, this PR takes in the new `lds_cache_data` column in `compute_kernel_roofline_data`, maps it through and exposes LDS as a roofline bw/intensity option in the compute UI.

## Test Plan

Generated a fresh `occupancy` workload analysis DB from rocprofiler-compute with:

- `schema_version = 1.4.0`
- `compute_kernel_roofline_data.lds_cache_data`
- non-zero LDS intensity values for applicable kernels

And ran roc-optiq compute data-model and controller tests against that DB.

## Test Result

- `datamodel-compute-tests`: passed
- `roc-optiq-controller-compute-tests`: passed

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -195,7 +196,10 @@ namespace DataModel
 			query += " total_flops, ";
 			query += " l1_cache_data, ";
 			query += " l2_cache_data, ";
-			query += " hbm_cache_data ";
+			query += " hbm_cache_data, ";
+			query += IsVersionGreaterOrEqual("1.4.0")
+				? " lds_cache_data "
+				: " 0 AS lds_cache_data ";
```
jamessiddeley-amd: Note on the schema-version guard here:

rocprofiler-compute only adds `compute_kernel_roofline_data.lds_cache_data` starting with compute DB schema `1.4.0`. Older schema `1.3.0` DBs do not have that column, so always selecting `lds_cache_data` would break for older DBs.

So this query gates on versions 1.4.0 or greater

- schema >= `1.4.0`: select the real `lds_cache_data`
- schema < `1.4.0`: return `0 AS lds_cache_data` so the query shape stays compatible with optiq

Worth mentioning as I'm not 100% sure whether Optiq generally aims to keep compatibility with older compute DB schemas or if it only needs to support the most recent schema from rocprofiler-compute.

I kept the fallback here to avoid breaking older DBs but happy to simplify this to always select `lds_cache_data` if Optiq only supports current schema versions.

src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -42,6 +42,7 @@ namespace DataModel
 		{"l1_cache_data", kRPVComputeColumnRooflineL1CacheData},
 		{"l2_cache_data", kRPVComputeColumnRooflineL2CacheData},
 		{"hbm_cache_data", kRPVComputeColumnRooflineHBMCacheData},
+		{"lds_cache_data", kRPVComputeColumnRooflineLDSCacheData},
```
drchen-amd: Intention of this list is to confirm presence of values that are expected to exist irrespective of version. Along with suggestion below, I would just not include LDS here.

src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -195,7 +196,10 @@ namespace DataModel
 			query += " total_flops, ";
 			query += " l1_cache_data, ";
 			query += " l2_cache_data, ";
-			query += " hbm_cache_data ";
+			query += " hbm_cache_data, ";
```
drchen-amd: Prefer to omit instead of 0 when it does not exist.
```suggestion
			query += " hbm_cache_data ";
			if (IsVersionGreaterOrEqual("1.4.0"))
			{ 
				query += ", lds_cache_data ";
			}
```

drchen-amd (commented):
Looks good, just some minor nitpicks. Also please target dev branch instead of main.

src/controller/src/compute/rocprofvis_controller_trace_compute.cpp
```diff
@@ -523,6 +523,7 @@ rocprofvis_result_t ComputeTrace::LoadRocpd(Future* future)
                                                     { kRPVComputeColumnRooflineL1CacheData, std::nullopt },
                                                     { kRPVComputeColumnRooflineL2CacheData, std::nullopt },
                                                     { kRPVComputeColumnRooflineHBMCacheData, std::nullopt },
+                                                    { kRPVComputeColumnRooflineLDSCacheData, std::nullopt },
```
drchen-amd:
```suggestion
```

drchen-amd (approved):

PR #867, opened 2026-06-08
Duplicate File Open Protection

Sajeeth-Wimalasuriyan:
Screenshot 2026-06-08 at 1 09 54 PM

- File > Open rejected a second file (.db or .rpv) whose underlying trace .db was already open in another tab, showing a confusing "<path> already open" toast and leaving the user on the current tab.

- Root cause: a project is identified by its underlying trace path, so any file that resolves to the same .db (a .db and a .rpv saved from it, or two .rpv files from one .db) collided — and the raw-path comparison made detection inconsistent.

-We keep the intended one-tab-per-trace behavior but make it correct and clear: the trace path is canonicalized before the duplicate lookup so every file resolving to an open trace is caught, and the toast is replaced with a popup that names the existing tab and switches to it.

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -952,6 +952,7 @@ AppWindow::RenderViewMenu(Project* project)
                 tool_bar_item->m_visible = settings.show_toolbar;
             }
         }
+#ifndef __APPLE__
         if(ImGui::MenuItem("Fullscreen", "F11", m_is_fullscreen))
         {
             if(m_notification_callback)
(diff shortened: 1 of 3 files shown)


src/view/src/rocprofvis_project.cpp
```diff
@@ -183,14 +163,15 @@ Project::OpenProject(std::string& file_path)
 Project::OpenResult
 Project::OpenTrace(std::string& file_path)
 {
-    OpenResult result    = Failed;
-    Project*   duplicate = AppWindow::GetInstance()->GetProject(file_path);
+    OpenResult result = Failed;
+    // canonicalize so a .db and the path stored in a .rpv resolve to the same trace
+    file_path = std::filesystem::weakly_canonical(file_path).string();
```
tomk-amd, line 168: use a different name for "result" the variable gets shadowed below, causing warning..

src/view/src/rocprofvis_appwindow.cpp
tomk-amd, line 961: Is this supposed be part of this PR?

tomk-amd (changes requested):
Fix the warning please

tomk-amd (approved):

PR #868, opened 2026-06-09
Whitespace Optimization 

Sajeeth-Wimalasuriyan:
Screenshot 2026-06-09 at 8 50 39 AM

### DIFF
--- a/src/view/src/rocprofvis_timeline_view.cpp
+++ b/src/view/src/rocprofvis_timeline_view.cpp
@@ -1726,18 +1726,14 @@ TimelineView::CalculateTrackCounts()
                 break;
         }
     }
+
+    BuildTrackCountLabels();
 }
 
 void
-TimelineView::RenderTrackStats(float available_width)
+TimelineView::BuildTrackCountLabels()
 {
-    constexpr float PAD_X         = 12.0f;
-    constexpr float PAD_Y         = 6.0f;
-    const float     content_width = std::max(0.0f, available_width - (2.0f * PAD_X));
-
-    FontManager& fonts = m_settings.GetFontManager();
-
-    const std::string total_label =
+    m_track_counts.total_label =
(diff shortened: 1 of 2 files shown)


src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -1684,6 +1684,136 @@ TimelineView::MakeGraphView()
...
+                break;
+            case TrackInfo::TrackType::Stream:
+                ++m_track_counts.streams;
+                break;
+            case TrackInfo::TrackType::Counter:
+                ++m_track_counts.counters;
+                break;
+            default:
+                ++m_track_counts.other;
+                break;
```
tomk-amd, line 1726: The track counts are only calculated once, when the trace is loaded?  If so why not make the label strings here, instead of recreating each frame as below..

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -1684,6 +1684,136 @@ TimelineView::MakeGraphView()
...
+        out += std::to_string(count) + " " + (count == 1 ? singular : plural);
+    };
+
+    std::string breakdown;
+    append_part(breakdown, m_track_counts.instrumented_threads, "thread", "threads");
+    append_part(breakdown, m_track_counts.sampled_threads, "sampled", "sampled");
+    append_part(breakdown, m_track_counts.queues, "queue", "queues");
+    append_part(breakdown, m_track_counts.streams, "stream", "streams");
+    append_part(breakdown, m_track_counts.counters, "counter", "counters");
+    append_part(breakdown, m_track_counts.other, "other", "other");
```
tomk-amd: Move to CalculateTrackCounts()

tomk-amd (changes requested):

tomk-amd (approved):

PR #869, opened 2026-06-09
Queue Utilization V3

drchen-amd:
Add support for tracks with multiple levels. This requires processing each event and is less performant than V2 (1847.11 ms vs 1504.86 ms for 2 million events). Number of operations in V3 scales linearly with number of events on the track.

  ## DM
  - Slice timestamp hashing:
    - Added enum parameter `rocprofvis_dm_hashed_timestamp_tag_t` to `Track::AddSlice`, `Track::GetSliceAtTime`, `Trace::CheckSliceExists`. This allows slices with identical timestamps to remain unique and not interfere with each other when their tags differ. All use cases except analysis uses tag `kRocProfVisDmHashedTimestampTagTrackSlice`.
    - Added typedef `rocprofvis_dm_hashed_timestamp_t` for functions that expect a hashed timestamp instead of a raw timestamp. Previously both the hash-combined timestamp and the raw timestamp were passed as `rocprofvis_dm_timestamp_t`, so it was not immediately clear from the signature whether a parameter should be hashed without checking the implementation.
    - Unified multiple identical implementations of hash_combine into `rocprofvis_dm_hash_combine_timestamp` to be shared with controller.
  - Moved slice query sorting out of SQL into C++ (without this, V3 would be 3269.87 ms).

  ## Controller
  - Updated for `rocprofvis_dm_hash_combine_timestamp` changes.
  - V3 queue utilization:
    - Calculates the union of all event ranges divided by the analysis rrange. Should work for any track.
    - Issues one query for entire range. Experimented with `TableProcessor` scheme of splitting into sub ranges over available threads to process in parallel but observed repeatable ~200ms performance regression. 

  ## View
  - Align and sync queue utilization area with counter scale area.
  - Fix queue utilization intermittently double-fetching the very first request due to incorrect initial state.
  - Fix assert from track meta area when track height is made small / font size is large.
  - Fix track name in meta area eliding prematurely.
  - Fix crash when opening a nonexistent file from File > Recent Files.

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -942,7 +942,7 @@ AppWindow::RenderFileMenu(Project* project)
             SettingsManager::GetInstance().GetInternalSettings().recent_files;
         if(ImGui::BeginMenu("Recent Files", !recent_files.empty()))
         {
-            for(const std::string& file : recent_files)
+            for(std::string file : recent_files)
             {
                 if(ImGui::MenuItem(file.c_str(), nullptr))
                 {
--- a/.agents/DATABASE.md
+++ b/.agents/DATABASE.md
@@ -172,7 +172,7 @@ database.
 rocprofvis_dm_result_t rocprofvis_db_read_metadata_async(database, future);
 rocprofvis_dm_result_t rocprofvis_db_cleanup_async(database, future, rebuild);
 rocprofvis_dm_result_t rocprofvis_db_read_trace_slice_async(
-    database, start_ns, end_ns, num_tracks, track_id_array, future);
+    database, start_ns, end_ns, hashed_timestamp_tag, num_tracks, track_id_array, future);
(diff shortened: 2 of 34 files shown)


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -942,7 +942,7 @@ AppWindow::RenderFileMenu(Project* project)
             SettingsManager::GetInstance().GetInternalSettings().recent_files;
         if(ImGui::BeginMenu("Recent Files", !recent_files.empty()))
         {
-            for(const std::string& file : recent_files)
```
tomk-amd, line 945: Is this for? :

Fix crash when opening a nonexistent file from File > Recent Files.

tomk-amd (approved):

src/view/src/rocprofvis_appwindow.cpp
drchen-amd: Yes. The open() will delete the file from settings manager list that this loop is referencing. When I originally added the recent files feature, this was a copy.

PR #871, opened 2026-06-11
Scope track tooltip to name label and move track options to context menu

dhingora-amd:
## Motivation

The track metadata tooltip previously appeared anywhere over the meta area, which was distracting during normal navigation. Track options were also tucked behind a gear button that wasn't very discoverable. This change makes the tooltip less intrusive and consolidates track actions into a single right-click menu.

## Technical Details

- The metadata tooltip now only shows when hovering the track **name label** (gated on the label's rect) instead of the entire meta area.
- Removed the gear button and moved `RenderMetaAreaOptions()` into the track's right-click context menu as a "Track Options" submenu.
- Context menu entries now have icons; added a reusable `IconBeginMenu` helper (alongside the existing `IconMenuItem`) since ImGui menu labels render in the text font, which lacks the icon glyphs.
- Reclaimed the gear's space for the track name label.

### DIFF
--- a/src/view/src/rocprofvis_track_item.cpp
+++ b/src/view/src/rocprofvis_track_item.cpp
@@ -23,6 +23,8 @@ inline constexpr float    DEFAULT_GRIP_WIDTH             = 20.0f;
 inline constexpr uint64_t DEFAULT_CHUNK_DURATION         = TimeConstants::ns_per_s * 30;
 inline constexpr float    META_TOOLTIP_MAX_WIDTH         = 320.0f;
 inline constexpr uint64_t META_TOOLTIP_COMPACT_COUNT_MIN = 1000;
+inline constexpr float    NAME_LABEL_HITBOX_PADDING_X    = 4.0f;
+inline constexpr float    NAME_LABEL_HITBOX_PADDING_Y    = 3.0f;
 constexpr const char*     TRACK_COPY_MENU_POPUP_NAME     = "TrackCopyMenu";
 
 float TrackItem::s_metadata_width = 400.0f;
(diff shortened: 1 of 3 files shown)


drchen-amd:
The hover detection is not responsive. Seems like mouse being inside does not register, but after being inside the mouse has to move for it to register.

[Screencast from 06-11-2026 01:25:47 PM.webm](https://github.com/user-attachments/assets/e409aae4-52ac-4064-93e5-6defcc554ca7)

> _Track options were also tucked behind a gear button that wasn't very discoverable._

I don't think moving them into a context menu makes them any more discoverable. If anything I think its less discoverable because the gear hints at options being available. Just my two cents, lets see what others think.

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -359,9 +346,13 @@ TrackItem::RenderMetaArea()
                                    !ImGui::IsAnyItemHovered() &&
                                    !m_pill.WasLastHovered();
 
+    const bool name_label_hovered =
+        meta_area_hovered && name_label_visible &&
+        ImGui::IsMouseHoveringRect(name_label_min, name_label_max);
+
```
tomk-amd, line 355: I would prefer if the tool tip showed up only when the actual text label is hovered, not the entire span of the "row".  It feels strange for the tool tip to activate on white space.

tomk-amd:
> I don't think moving them into a context menu makes them any more discoverable. If anything I think its less discoverable because the gear hints at options being available. Just my two cents, lets see what others think.

I would prefer to bury the track options in the context menu.  These options are not frequently used and the interface looks cleaner without a the icon on every track.

tomk-amd:
image

One thing that I think we should do, via another PR later is rework, how queue utilization is displayed.. 

This looks messy when the expand arrow is needed.  Move utilization out of the RenderMetaAreaScale() function, but where? Any suggestions welcome.

drchen-amd:
I think we could change the pill into a horizontal list of pills. The queue utilization could go in a pill and so could other stuff in the future like the comparison markings. The pills would hide/show depending on the available width.

tomk-amd (approved):

PR #872, opened 2026-06-11
Add icons to context menus across the view

dhingora-amd:
## Motivation

Right-click context menus throughout the app used plain text menu items, while the compute metric tables already showed icons next to each action. This PR brings the rest of the menus in line so the context-menu styling is consistent across the whole view layer, making actions quicker to scan and visually matching the compute tables.

## Technical Details

Converted the remaining hand-rolled and `ImGui::MenuItem`-based context menus to use the shared `IconMenuItem` helper, with context-appropriate glyphs:

- **Shared table row menu** (`InfiniteScrollTable`) — used by the event, sample, event-search, and kernel-instance tables: compass for "Go To Event/Sample", copy for copy actions, archive for export.
- **Timeline event menu** (`TimelineView`): grid for time-range selection, copy for copy actions, trash for removal, note for annotation, compass for navigation.
- **Event call-stack and flow menus** (`EventsView`): compass and copy. These were hand-rolled separately from the table menu, so they were updated to match.
- **Kernel bar-chart column menu** (`KernelMetricTable`): bar-chart icon.

`IconMenuItem` gained an optional `enabled` parameter so it can fully replace `ImGui::MenuItem` in menus that have conditionally disabled entries. Redundant explicit `ImGui::CloseCurrentPopup()` calls were removed since `IconMenuItem` already closes the popup on click.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
+++ b/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
@@ -924,7 +924,7 @@ KernelMetricTable::RenderBarChartContextMenu(int col)
     if(ImGui::BeginPopupContextItem("##bar_ctx"))
     {
         bool has_bars = m_bar_chart_columns.count(col) > 0;
-        if(ImGui::MenuItem(has_bars ? "Hide Bar Chart" : "Show Bar Chart"))
+        if(IconMenuItem(ICON_CHART_BAR, has_bars ? "Hide Bar Chart" : "Show Bar Chart"))
         {
             if(has_bars)
             {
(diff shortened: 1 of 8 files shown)


drchen-amd:
image
image


I'm not a fan of the misaligned of the text starting positions and the large vertical padding. It would look more clean if all text items are aligned to the same starting position and each menu item maintained the standard height of one frame with or without icon.

tomk-amd:
> I'm not a fan of the misaligned of the text starting positions and the large vertical padding. It would look more clean if all text items are aligned to the same starting position and each menu item maintained the standard height of one frame with or without icon.

I agree the menu items' text should be aligned.  I think IconMenuItem() should be improved to enforce this.  This function is a relatively new addition to the project so it can be improved.

tomk-amd (changes requested):
Please modify IconMenuItem to fix the horizontal alignment.  The Icon "column" will should be set to some reasonable fixed width. (Remember to account for the font size potentially changing).

Also the bottom vertical padding between items is too large.

tomk-amd (approved):

drchen-amd:
Regression from this change: 
Track options icon does not display when options are showing.
image

PR #873, opened 2026-06-11
macos mission control fix

Sajeeth-Wimalasuriyan:
On macOS, system gestures (e.g. triggering Mission Control via Ctrl+Up while dragging the window to a new Space, screenshot chords, etc.) can consume a modifier key-up before GLFW observes it. This leaves GLFW's cached key state stuck "down". Because ImGui enables ConfigMacOSXBehaviors on macOS, a phantom-stuck Control key causes every left-click to be translated into a right-click, so buttons and menus stop responding.

This PR recovers from that state by reading the live OS modifier state directly (independent of GLFW's cached per-window state) and feeding it back into ImGui to clear any phantom modifier.


tomk-amd (approved):
Please move rocprofvis_platform_helpers.h into the src folder.  The inc/ folder is only for public interfaces, the helper files does not need to be public.

PR #876, opened 2026-06-15
Meta Area Cleanup

Sajeeth-Wimalasuriyan:
image

### DIFF
--- a/src/view/src/rocprofvis_flame_track_item.cpp
+++ b/src/view/src/rocprofvis_flame_track_item.cpp
@@ -782,13 +782,6 @@ FlameTrackItem::RenderChart(float graph_width)
 }
 
 void
-FlameTrackItem::RenderMetaAreaScale()
-{
-    // Queue utilization is drawn as a pill beside the QUEUE pill in
-    // RenderSecondaryMetaPill, so no separate scale column is rendered here.
-}
-
-void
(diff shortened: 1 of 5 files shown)


src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -784,48 +784,52 @@ FlameTrackItem::RenderChart(float graph_width)
...
-        ImGui::EndDisabled();
-        ImGui::GetWindowDrawList()->AddLine(
-            ImVec2(window_pos.x + content_region.x - m_meta_area_scale_width,
-                   window_pos.y),
-            ImVec2(window_pos.x + content_region.x - m_meta_area_scale_width,
-                   window_pos.y + content_region.y),
-            m_settings.GetColor(Colors::kMetaDataSeparator), SCALE_SEPERATOR_WIDTH);
-    }
+    // Queue utilization is drawn as a pill beside the QUEUE pill in
+    // RenderSecondaryMetaPill, so no separate scale column is rendered here.
```
tomk-amd: This comment feels a bit too specific to the current task.  I think it should just say //Nothing to render, or make the base class implementation non-abstract and don't implement here.

tomk-amd (approved):
Looks good except for the comment

PR #877, opened 2026-06-15
Make Event Details table right-click hitboxes consistent and dedupe scaffolding

dhingora-amd:
## Motivation

In the Event Details panel, right-click behavior was inconsistent across
tables. The Flow and Call Stack tables opened their context menu when you
right-clicked anywhere in a cell, but the Arguments table only responded
when you clicked directly on the cell's text, not the surrounding cell
area. On top of that, right-clicking directly on the text in the Basic
Data, Extended Data, and Arguments tables failed to open the menu at all.

This PR makes the right-click hitbox span the full row for every Event
Details table so the interaction is predictable and consistent, and fixes
the missed right-clicks on cell text.

## Technical Details

- Made the right-click hitbox span the full row (rather than just the
  text bounds) across all Event Details tables: Basic Data, Extended
  Data, Arguments, Flow, and Call Stack.
- Fixed a bug where right-clicking directly on a cell's text did not open
  the context menu in the Basic Data, Extended Data, and Arguments tables
  by adding per-cell right-click capture.
- Refactored the duplicated context-menu scaffolding into shared helpers
  (row hit-box rendering, per-cell right-click capture, styled popup
  begin/end, and copy-menu item insertion) to remove repetition across
  the five tables.
- Replaced the multiple member variables that tracked context-menu
  row/column state with a single `CellMenuTarget` struct.
- Minor cleanup: removed unused constants and includes, and standardized
  on `COPY_DATA_NOTIFICATION` for the copy toast.

### DIFF
--- a/src/view/src/rocprofvis_events_view.cpp
+++ b/src/view/src/rocprofvis_events_view.cpp
@@ -11,6 +11,9 @@
 #include "rocprofvis_utils.h"
 #include "widgets/rocprofvis_notification_manager.h"
 
+#include <array>
+#include <vector>
+
 namespace RocProfVis
 {
 namespace View
@@ -23,6 +26,9 @@ constexpr ImGuiTableFlags TABLE_FLAGS = ImGuiTableFlags_BordersOuter |
 
 constexpr int kFlowColumnCount      = 6;
 constexpr int kCallStackColumnCount = 5;
+constexpr int kArgColumnCount       = 4;
+constexpr int kBasicColumnCount     = 2;
+constexpr int kExtColumnCount       = 2;
 
 namespace
 {
@@ -63,6 +69,12 @@ EventsView::EventsView(DataProvider&                      dp,
 , m_context_menu_flow_column(-1)
 , m_context_menu_callstack_index(-1)
 , m_context_menu_callstack_column(-1)
+, m_context_menu_arg_index(-1)
+, m_context_menu_arg_column(-1)
+, m_context_menu_basic_index(-1)
+, m_context_menu_basic_column(-1)
+, m_context_menu_ext_index(-1)
+, m_context_menu_ext_column(-1)
 {
     static_assert(CallStackHoverState::kInvalidId ==
                       TimelineSelection::INVALID_SELECTION_ID,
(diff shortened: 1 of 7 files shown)


src/view/src/rocprofvis_events_view.cpp
```diff
@@ -37,6 +44,104 @@ PushSectionHeaderStyle(SettingsManager& settings)
...
+}
+
+// Renders the transparent, full-row selectable that acts as the right-click
+// hit-box for the empty areas of a table row. Cells drawn on top capture
+// right-clicks that land directly on their text (see CaptureCellRightClick).
+// On a right-click, records the row and hovered column into target and raises
+// open. Returns whether the row is hovered (for double-click / highlight use).
+bool
+RenderRowHitbox(const char* hitbox_id, int row, int column_count,
+                CellMenuTarget& target, bool& open)
```
drchen-amd: What is our opinion on these unnamed namespaces? We used to not have these, but AI loves and I've been seeing it more and more. 

It would be nice if this CellMenuTarget was private, but these functions are outside of class name space and won't have access.

drchen-amd (commented):
>Minor cleanup: removed unused constants and includes, and standardized
on COPY_DATA_NOTIFICATION for the copy toast.

I'm not sure if this is intentional but none of the copies show the notification anymore.

src/view/src/rocprofvis_events_view.cpp
tomk-amd: Yup AI loves these.  I don't have a problem with the anonymous namespace in these cases as this is the "modern" way of having private free floating helper functions.  

My 2 cents:

Perhaps some of these helpers could be re-used, in this case they can be moved to a utils file and made available to everyone.

If truly private I think these functions could be added as private static members of the calling class? This would allow CellMenuTarget to be private too.  The larger question is can the functions and CellMenuTarget be re-used else where in the app?

tomk-amd:
> I'm not sure if this is intentional but none of the copies show the notification anymore.

Not intentional, this notification should be restored.

src/view/src/widgets/rocprofvis_gui_helpers.cpp
```diff
@@ -565,5 +567,96 @@ DrawInternalBuildBanner(const char* text /*= "Internal Build"*/)
...
+AddCopyRowCellMenuItems(const std::string* cells, int column_count, int column)
+{
+    if(IconMenuItem(ICON_COPY, "Copy Row Data"))
+    {
+        std::string row_text;
+        for(int c = 0; c < column_count; c++)
+        {
+            if(c > 0) row_text += '\t';
+            row_text += cells[c];
+        }
```
tomk-amd, line 788: This notification should say; ""Row data copied to clipboard", like in infinite scroll table.

A const for this string is not defined. But we can add one.

The definition for COPY_DATA_NOTIFICATION maybe should move to gui_helpers too.

I think these three functions might as well move to gui_helpers too:

```
bool
CopyableTextUnformatted(
    const char* text, std::string_view unique_id = "", std::string_view notification = "",
    bool one_click_copy = false, bool context_menu = false,
    std::function<void(const char* value_to_copy)> menu_func = nullptr);

bool IconMenuItem(const char* icon, const char* label, bool enabled = true);

bool IconBeginMenu(const char* icon, const char* label);
```

tomk-amd: And while we are cleaning up, I think 

```
void
WithPadding(float left, float right, float top, float bottom,
            const std::function<void()>& content);
```

can be removed.  This is unused as far as I can tell.

tomk-amd (approved):

PR #882, opened 2026-06-16
Remote file access and profiler launching

tomk-amd:
## Motivation

This PR adds three in-development features plus supporting infrastructure:

- **Remote access (SSH):** browse a remote filesystem, download files, store SSH
  connection profiles, host-key + interactive auth prompts.
- **Profiler launch:** configure and launch a profiler locally.
- **Remote profiler launch:** launch a profiler over SSH and pull the trace back.
- **Operation Monitor (`AppMonitor`):** tracks long-running operations (profiler
  sessions, SSH ops) and surfaces them in the status bar / shutdown drain.

The features are **not production-ready**; they ship gated behind CMake feature
flags that **default OFF**, so a normal build/release is unchanged. 

#### 1.1 Feature flags (CMake)

Two options, both `OFF` by default, defined in the root `CMakeLists.txt`:

| Option | Default | Enables |
|--------|---------|---------|
| `ROCPROFVIS_ENABLE_PROFILER` | `OFF` | In-app profiler launch UI + backends + the profiler controller/ABI |
| `ROCPROFVIS_ENABLE_REMOTE`   | `OFF` | Remote/SSH UI + controller SSH stack + the `libssh2`/`mbedtls` (or OpenSSL) dependency |

They are **independent**:

- Profiler launch builds without remote (the SSH connection UI inside the
  launcher is compiled out via `ROCPROFVIS_ENABLE_REMOTE` guards).
- Remote can be enabled on its own.
- The **remote-profiling glue** (`RemoteProfilerSession` / `SshProfilerExecutor`)
  is only built when **both** flags are ON.

`AppMonitor` and the small JSON/profile helpers are **always** compiled (they are
generic infra woven into the app lifecycle), but they are inert without the
features.

Enable them at configure time on top of a preset, e.g.:

```bash
# Local dev (Linux release preset) with both features + default mbedTLS crypto
cmake --preset linux-release -DROCPROFVIS_ENABLE_PROFILER=ON -DROCPROFVIS_ENABLE_REMOTE=ON
cmake --build build/linux-release
```

```powershell
# Windows
cmake --preset x64-debug -DROCPROFVIS_ENABLE_PROFILER=ON -DROCPROFVIS_ENABLE_REMOTE=ON
```

For convenience a CMakeUserPresets file is attached that has these configurations: 
[CMakeUserPresets.json](https://github.com/user-attachments/files/29910740/CMakeUserPresets.json)

#### 1.2 Dependencies

The remote feature pulls in git submodules `thirdparty/libssh2` and `thirdparty/mbedtls` — run `git submodule update --init --recursive` before configuring with `ROCPROFVIS_ENABLE_REMOTE=ON`.

See the **Crypto backend (SSH / remote features)** section in the BUILDING.md doc for instructions on how to get the OpenSSL dependency.

#### 1.3 Crypto backend for remote (libssh2)

libssh2 needs a crypto backend. Selected with the `CRYPTO_BACKEND` cache
variable (only consulted when `ROCPROFVIS_ENABLE_REMOTE=ON`):

| `CRYPTO_BACKEND` | Default | How it's provided |
|------------------|---------|-------------------|
| `mbedTLS` | ✅ default | **Vendored** submodule (`thirdparty/mbedtls`), built statically. No system dependency. |
| `OpenSSL` | opt-in  | **External** dependency resolved via `find_package(OpenSSL REQUIRED COMPONENTS SSL Crypto)`. Not vendored. |

```bash
# Remote build using OpenSSL instead of the default mbedTLS
cmake --preset linux-release \
  -DROCPROFVIS_ENABLE_REMOTE=ON \
  -DCRYPTO_BACKEND=OpenSSL
```

OpenSSL notes:
- Provide `OPENSSL_ROOT_DIR` if it isn't auto-found (e.g. Homebrew
  `openssl@3` on macOS, or a Windows OpenSSL install root).
- On **macOS**, the build stages the resolved `libssl`/`libcrypto` dylibs into
  the app bundle and rewrites their install names to `@rpath/...`
  (`cmake/macos_openssl_fixup.sh.in`).
- On **Windows**, the resolved `libssl`/`libcrypto` DLLs are copied next to the
  executable post-build.
- The default **mbedTLS** path needs none of this (statically linked, no runtime
  staging).

#### 1.4 CI runners

**CI defaults everything for the new features OFF** — normal/nightly/release runs
are unaffected. Three `workflow_dispatch` inputs let a run opt in:

| Input | Type | Default | Effect |
|-------|------|---------|--------|
| `enable_remote`   | boolean | `false`   | `-DROCPROFVIS_ENABLE_REMOTE=ON` |
| `enable_profiler` | boolean | `false`   | `-DROCPROFVIS_ENABLE_PROFILER=ON` |
| `crypto_backend`  | string  | `mbedTLS` | `-DCRYPTO_BACKEND=<value>`; each runner installs OpenSSL only when set to `OpenSSL` (e.g. macOS `brew install openssl@3` + `OPENSSL_ROOT_DIR`) |

`crypto_backend` only has an effect when `enable_remote=true` (the whole
libssh2/crypto stack is guarded by `if(ROCPROFVIS_ENABLE_REMOTE)`).

Continuous-integration workflows (`workflow_dispatch`) and their names:

| Workflow name | File |
|---------------|------|
| `Mac Continuous Integration` | `.github/workflows/ci-mac.yml` |
| `Windows Continuous Integration` | `.github/workflows/ci-windows.yml` |
| `Ubuntu Continuous Integration` | `.github/workflows/ci-ubuntu.yml` |
| `RedHat (Rockylinux) Continuous Integration` | `.github/workflows/ci-redhat.yml` |

Release/package workflows read the backend from the `repository_dispatch`
`client_payload.crypto_backend` (default `mbedTLS`):
`macos-release-build-package.yml`, `windows-release-build-package.yml`,
`ubuntu-release-build-package.yml`, `redhat-release-build-package.yml`,
`oracle-release-build-package.yml`.

Example — build & test the **full** feature set (remote + profiler) with the
OpenSSL backend on macOS, uploading artifacts:

```bash
gh workflow run "Mac Continuous Integration" \
  --ref <branch> \
  -f enable_remote=true \
  -f enable_profiler=true \
  -f crypto_backend=OpenSSL \
  -f upload_to_github=true
```

```bash
# Windows / Ubuntu / RedHat equivalents
gh workflow run "Windows Continuous Integration" --ref <branch> -f enable_remote=true -f enable_profiler=true -f crypto_backend=OpenSSL -f upload_to_github=true
gh workflow run "Ubuntu Continuous Integration"  --ref <branch> -f enable_remote=true -f enable_profiler=true -f crypto_backend=OpenSSL -f upload_to_github=true
gh workflow run "RedHat (Rockylinux) Continuous Integration" --ref <branch> -f enable_remote=true -f enable_profiler=true -f crypto_backend=OpenSSL -f upload_to_github=true
```

```bash
# Remote with the default (vendored) mbedTLS backend — omit crypto_backend
gh workflow run "Ubuntu Continuous Integration" --ref <branch> -f enable_remote=true -f enable_profiler=true
```

The release/package workflows read `crypto_backend` from the `repository_dispatch` `client_payload`; if a remote-enabled release build is ever needed, the `enable_remote`/`enable_profiler` flags would need to be plumbed there too (currently CI-only, by design — releases ship the features OFF).

---

### 2. Design overview

This is a high-level tour of the three features and how they compose. Everything
follows the existing app layering — **view → controller → model** — and all the
new long-running work is asynchronous: the controller does the blocking work on
its `JobSystem` thread pool behind `Future`s, and the view stays single-threaded
(ImGui), polling status once per frame and reacting to events.

#### 2.1 The shared spine: Operation Monitor (`AppMonitor`)

`AppMonitor` is a small, generic, always-compiled registry of "long-running
operations." It doesn't know what SSH or profiling is — each operation just
supplies a few callbacks:

- a **status poller** (read once per frame),
- an **event factory** (emit a typed event when the status changes),
- a **terminal test**, and optionally a **cancel** hook, a bound `Future`, and
  **teardown** closures.

Each frame it polls every operation, queues a status-changed event on the
`EventManager`, and reaps terminal/cancelled operations — but only after their
`Future` resolves, running teardown closures then (deferred, non-blocking). This
is what lets a session be destroyed while work is still in flight without
freeing resources out from under a worker. It also feeds the status bar
("Background: N operations") and gates app shutdown until work drains.

The two new event types it emits are `kProfilerStatusChanged` and
`kRemoteStatusChanged`.

#### 2.2 Remote access (SSH)

**Controller side** owns the actual SSH via `libssh2`:

- `SshConnection` — one connection: a libssh2 session + socket + an `SshBridge`.
- `SshClient` — allocates connections and runs the blocking operations
  (connect, authenticate, execute, download, browse directory).
- `SshBridge` — the worker↔UI rendezvous. Workers push status / stdout / file
  listings into it and block on it for interactive input (host-key decisions,
  password / keyboard-interactive prompts) via a condition variable. It's a
  controller `Handle`, so the view reads all of this over the C ABI.
- `KnownHosts` — OpenSSH `known_hosts` load/check/add (host-key verification).
- `Remote` — the C-ABI dispatch layer; each action is a `JobSystem` job bound to
  a `Future`.

**View side** drives the workflow:

- `SshSession` — owns a connection handle and runs one phase at a time
  (connect → auth → …); each phase registers an `AppMonitor` operation whose
  poller reads the bridge status and emits `kRemoteStatusChanged`.
- `RemoteTraceOrchestrator` — a phase state machine for the "open a remote trace"
  flow (connect → auth → browse/download → open).
- `RenderSshAuthModal` — the host-key and keyboard-interactive / password popups,
  fed by snapshot types (`PromptRequest` / `HostKeyRequest`) the session exposes.
- `SshConnectionStore` / `SshConnectionConfig` — saved connection profiles
  (persisted in `profiles.json`); `RemoteUri` carries the live connection + path
  state through a workflow.

Flow (all non-blocking, event-driven on the main thread): start a phase → a
`JobSystem` worker does it → the bridge status advances → `AppMonitor` emits
`kRemoteStatusChanged` → the orchestrator/session advances to the next phase (or
the auth modal pops for a prompt/host-key). Download progress surfaces through
`FileStat` snapshots.

#### 2.3 Profiler launch (local)

**Controller side:**

- `ProfilerConfig` — the launch spec (profiler path, target, args, env,
  connection type).
- `ProfilerProcessController` — the run state machine (`Idle → Running →
  Completed/Failed/Cancelled`); owns an `IProfilerExecutor`.
- `LocalProfilerExecutor` — spawns the process (`fork`/`execvp` on POSIX,
  `CreateProcess` on Windows), captures stdout/stderr, reaps on exit/cancel.
- The profiler **C ABI** (`rocprofvis_profiler_*`) wraps a `ProfilerSession`
  handle (controller + bound future); a `JobSystem` "monitor job" pumps output
  and state until the run ends.
- `Cmdline` — shared argv/env/command-string composition (POSIX shell, Windows
  command line, human-readable preview).

**View side:**

- `IProfilerBackend` / `RocprofSysBackend` — per-profiler knowledge: the config
  tabs, argv construction, and **parsing the produced trace path from the
  profiler's stdout** (the authoritative trace-path source). Adding a profiler
  (Compute, rocprofv3) means adding a backend.
- `ProfilerSessionBase` / `ProfilerSession` — own the C-ABI objects and register
  the profiler `AppMonitor` status poller (`kProfilerStatusChanged`).
- `ProfilerLaunchOrchestrator` — drives one run, normalizes its state/output for
  the UI, and on completion resolves the trace path (via the backend scraper) and
  optionally opens it.
- `ProfilerLauncherDialog` — the configuration + console UI; `LaunchConfig` +
  `LaunchPresetManager` persist launch presets.

#### 2.4 Remote profiler launch (the two combined)

Remote profiling is where the SSH and profiler stacks meet — built only when
**both** feature flags are on. `RemoteProfilerSession` (view) drives a sequential
phase machine that reuses both subsystems over **one** SSH connection:

```
Connect ──▶ Authenticate ──▶ Profile (remote) ──▶ Download trace ──▶ Open
  (SshSession /  kRemoteStatusChanged)   │        (SshSession SFTP,   (auto-load,
                                         │         kRemoteStatusChanged) if enabled)
                                         └─ profiler C ABI over the borrowed
                                             connection (SshProfilerExecutor),
                                             tracked via kProfilerStatusChanged
```

Step by step:

1. The launcher builds a `ProfilerConfig` tagged as an SSH launch and starts a
   `RemoteProfilerSession`, which opens an `SshSession` and begins **Connect**.
2. **Connect → Authenticate** run as SshSession phases; the auth modal surfaces
   host-key / credential prompts as needed. The connection is now live and idle.
3. **Profile:** `rocprofvis_profiler_launch_remote_async(...)` runs the profiler
   command on the borrowed connection via `SshProfilerExecutor`, streaming output
   into the bridge; the session tracks it with a profiler `AppMonitor` op.
4. On **Completed**, the session parses the remote trace path from the captured
   stdout (backend scraper) and starts **Download** (SFTP) on the same session.
5. On download completion it opens the local copy (honoring the "open trace when
   complete" option).

The whole sequence is non-blocking and main-thread-driven: each phase kicks off
controller `JobSystem` work, and phase transitions happen in event handlers as
`AppMonitor` reports status. Because the profiler worker borrows the SSH
connection, teardown is carefully ordered (deferred via `AppMonitor`) so the
connection outlives the worker.

#### 2.5 Cross-cutting patterns

- **Async:** controller `Future` + `JobSystem` for all blocking work; the view
  never blocks.
- **Status / events:** `AppMonitor` polls controller status each frame and emits
  typed `RocEvent`s; orchestrators/sessions react on the main thread.
- **Deferred teardown:** resources tied to an in-flight `Future` are freed only
  after it resolves, so cancel/close/shutdown never race a worker.
- **Feature-gated:** everything above is compiled out unless the relevant flag is
  set (§1), keeping default builds untouched.

### 3. Known issues / flaws

These features are **in development and not production-ready**, which is why they
ship gated OFF (§1). The items below are the **known remaining flaws** deliberately deferred past this merge.

#### 3.1 Security / privacy

- **Credentials stored in plaintext (highest priority).** SSH passwords and key
  passphrases are serialized as plain JSON into `{config}/profiles.json` (default
  file permissions), and held unscrubbed in memory. **Must** move to a per-OS
  secure store (Windows Credential Manager / macOS Keychain / Linux
  libsecret/keyring), or at minimum stop persisting secrets, before the remote
  feature graduates. Prefer key/agent auth over stored passwords.
- **Environment values logged in cleartext.** Two `info` log sites print env
  `NAME=VALUE` pairs, which can leak secrets passed as env vars. Redaction is deferred. This feature may not be necessary if no secret info is ever passed via environment vars.

#### 3.2 Architecture / correctness debt

- **Remote SSH exec runs on a raw `std::thread`.** `SshProfilerExecutor` spawns
  its own worker (outside the controller `JobSystem`), decoupled from the bound
  `Future`, over a non-thread-safe borrowed libssh2 session. The lifetime/
  deadlock hazards this caused are **already fixed**, but the structural debt
  remains. The intended cleanup is to run the remote exec **as a JobSystem job**
  so the future *is* the work (removing the join, the "wait for worker" loop, and
  some teardown plumbing). Best done as its own PR.

#### 3.3 Platform

- **`fork()` path is not macOS-safe.** The POSIX local-profiler launch uses
  `fork()`+`execvp()` with non-async-signal-safe work in the child; fine on
  Linux (the only POSIX target today), unsafe on macOS. Must be reworked to
  `posix_spawnp()` before enabling the profiler on macOS. A `NOTE (macOS)`
  comment marks the call site.

#### 3.4 Robustness

- **`WaitSocket` conflates cancel-polling with a network-dead deadline.** Its 10s
  `select` timeout doubles as a "give up" threshold, and callers treat a plain
  timeout as failure. Latent, not live: the current profiler is chatty over
  stdout so the SSH channel never idles that long. Would need the two roles
  decoupled if a genuinely silent long-running remote command becomes a use case.

#### 3.5 UX

- **The profiler launcher UI/UX needs a redesign.** The dialog grew organically
  and carries rough edges (translation-unit `static` ImGui buffers, a hand-rolled
  preset system/modals instead of the shared helpers, opaque backend-owned
  settings tabs, some hardcoded colors). The remaining Medium/Low launcher items
  in the review are best treated as inputs to that rework rather than individual
  spot-fixes.
- **Remote file browsing** This is implemented in a test ssh connection dialog. 
  A proper remote filesystem browsing UI needs to be designed.

#### 3.6 CI

- CI can build the features on demand (§1.3) but does **not** build them by
  default, and the release/package workflows do not expose the feature flags at
  all. Before graduation, decide the default CI coverage and whether releases
  ever ship the features enabled.

Attached is a document with more details about these issues:
[remote-profiler-deferred-task.md](https://github.com/user-attachments/files/29910685/remote-profiler-deferred-task.md)


src/controller/inc/rocprofvis_controller_enums.h
```diff
@@ -117,6 +119,14 @@ typedef enum rocprofvis_controller_object_type_t
...
     // Roofline object
     kRPVControllerObjectTypeRoofline = 104,
+    // Roofline object
+    kRPVControllerObjectTypeRemoteConnection = 105,
+    // Profiler configuration
+    kRPVProfilerConfig = 106,
+    // Profiler session (owns ProfilerProcessController)
+    kRPVProfiler = 107,
+    // Ssh intercomponent bridge
+    kRPVSshBridge = 108,
```
drchen-amd: Should not be in the compute ifdef.

src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -384,6 +410,26 @@ AppWindow::ShowOpenFileDialog(const std::string& title, const std::vector<FileFi
...
+        ShowNativeFileDialog({}, initial_path, callback, false, true);
+        return;
+    }
+#endif
+    m_file_dialog_callback = callback;
+    m_init_file_dialog     = true;
+    IGFD::FileDialogConfig config;
+    config.path = initial_path;
+    config.flags = ImGuiFileDialogFlags_Modal;
+    ImGuiFileDialog::Instance()->OpenDialog(FILE_DIALOG_NAME, title, nullptr, config);
```
drchen-amd: Can use AppWindow::ShowImGuiFileDialog.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -4966,5 +4966,9 @@ DataProvider::ProcessMetricPivotTable(RequestInfo& req)
 }
 #endif
 
+// Profiler launching has moved to RocProfVis::View::ProfilerSession
+// (src/view/src/profiler/rocprofvis_profiler_session.{h,cpp}), monitored via
+// the AppMonitor. DataProvider no longer owns any profiler state.
+
```
drchen-amd: This file should be reverted.

src/view/src/rocprofvis_data_provider.h
```diff
@@ -18,6 +18,7 @@
 #include <optional>
 #include <string>
 #include <unordered_map>
+#include <utility>
```
drchen-amd: This file should be reverted.

src/view/src/rocprofvis_appwindow.cpp
tomk-amd: This was intentional.. The interface ShowImGuiFileDialog() would need to change with a folder_mode=true flag or something like that.  Can (and probably should) be done later

tomk-amd: Refactored this now anyway, as path reading callback was broken anyways for folder mode.

drchen-amd (approved):
Tested the following:

- Existing functionality is unchanged.
- Able to build with ROCPROFVIS_ENABLE_REMOTE and ROCPROFVIS_ENABLE_PROFILER on Windows and Ubuntu 22.04 with default and OpenSSL backends.
- Able to transfer and open a trace file located on Ubuntu system from Windows system.
- Able to profile workload on Ubuntu system and have the output db automatically open.
- Able to profile workload on Ubuntu system remotely from Windows system and have the output db automatically open on Windows.

Also attaching claude's review:
[REVIEW.md](https://github.com/user-attachments/files/30026507/REVIEW.md)

PR #888, opened 2026-06-18
Make flame-chart events more compact and fix resize/expand behavior

Sajeeth-Wimalasuriyan:
image

Made flame-chart event tracks more compact by scaling the event row height to the active text font instead of a fixed 40px, vertically centering labels, and adding a 1px gap between stacked events. Enforced a per-row minimum track height so resizing can no longer squeeze the meta area small enough for its controls to collide. Finally, reworked the expand/contract arrow logic so a squished one- or two-level track reliably shows the expand arrow to reopen it.


tomk-amd:
In compact mode, the track gets too compact when the up (set min size) arrow gets pressed.

Expanded:

image

Minimized:

image

tomk-amd (approved):

drchen-amd:
FYI incase this behavior change is not intended...
When you toggle compact mode, new implementation will self expand while old implementation does not. This happens with some tracks but not all. You can see that the first track I try gets auto expanded, but the second track does not.

https://github.com/user-attachments/assets/f0dd8b38-4f51-4280-9245-922f08fbf2d1

(Left is this with this commit, right is without)

drchen-amd:
One more difference from this change. On my home display, I have to be at the minimum font size (12) to get the event names to show up. Prior to this change, I could use any font size and could see event names.

image

(Font size 13, Left is with this change, right is a5d7f78c99318c983c7777c0825b4c2c75bde07a)

Does not matter how tall I make the track or how much I zoom in. The labels do not show up unless I use minimum font size.

It is display specific because it does not happen when I use laptop's embedded display. I also did not notice during the week on the office displays where I do not use minimum font size.

tomk-amd:
> It is display specific because it does not happen when I use laptop's embedded display. I also did not notice during the week on the office displays where I do not use minimum font size.

Ok we should get to the bottom of this and fix it.  Just out of curiousity what is the resolution and is any scaling applied?

image

drchen-amd:
7680 x 2160, 150% scaling.

The problem is using the literal font size to scale non text elements. Typically, this is only used as parameter for `pushfont` or `addtext`. These functions internally apply scaling such that the size you pass is not the size on screen. This change uses to it to scale the event box; a rectangle that takes the size you give as absolute. There is a mismatch between the size the event box gives to label, and the actual size required by the label.

I will create a change and you guys can test it to see if something else breaks since I'm not familiar with the original ask of this pr. Levels will be slightly taller than it current is.

PR #892, opened 2026-06-18
Adopt IMGUI DPI to Replace Existing Solution

Sajeeth-Wimalasuriyan:
## Summary
Replaces the project's custom DPI handling with Dear ImGui's built-in (1.92+) per-monitor DPI font scaling, and turns the font-size preference into a simple user slider.
- **Removes the bespoke DPI machinery:** `rocprofvis_view_set_dpi`, the GLFW `content_scale_callback` / `glfwSetWindowContentScaleCallback` wiring, `SettingsManager::Get/SetDPI` + `m_display_dpi`, the `dpi_based_scaling` user setting (+ its JSON key), and `FontManager::GetDPIScaledFontIndex`.
- **Enables ImGui auto-DPI:** sets `io.ConfigDpiScaleFonts` / `io.ConfigDpiScaleViewports`, enables multi-viewport (`ImGuiConfigFlags_ViewportsEnable`) with the `UpdatePlatformWindows()` / `RenderPlatformWindowsDefault()` pump in the main loop, and `style.FontScaleMain = 1.0f`.
- **Font size is now a user slider** (`SettingsPanel`), clamped to a supported range; previously persisted indices are clamped on load for backward compatibility.
- **Drops manual `* dpi` multipliers** on fixed window sizes (shutdown dialog, minimap popup) and makes the internal-build banner viewport/font-scale aware instead of using `io.DisplaySize`.
- **Status bar height** now tracks the current font size via a single guarded check in `UpdateStatusBar()`, covering both user-driven and DPI-driven font changes (ImGui auto-DPI resizes fonts without firing `kFontSizeChanged`).


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -1493,6 +1484,15 @@ AppWindow::ShowImGuiFileDialog(const std::string& title, const std::vector<FileF
 void
 AppWindow::UpdateStatusBar()
 {
+    // Auto-DPI changes the font size without firing kFontSizeChanged, so
+    // resize the status bar slot whenever the font size changes.
+    const float font_size = ImGui::GetFontSize();
```
tomk-amd: Will this be true for anywhere that `RocEvents::kFontSizeChanged` is used to detect fontsize changes?

Sajeeth-Wimalasuriyan: It wasn't — the poll only covered the status bar. Now FontManager::Update() detects the font-size change once per frame and re-fires kFontSizeChanged, so all subscribers (status bar, TimelineView, Pill) are covered. AppWindow is back to a normal subscription. (36e3cb71)

amokiche-amd:
I think the font size slider should have a wider range of values. Some people might use the tool on a huge monitor with a lower resolution, while others might use it on a laptop. Right now, if I run the app on a laptop and set the font size to the maximum, I'm comfortable with it, but I can easily imagine people who might find the text too small.

src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -1493,6 +1484,15 @@
...
+    if(font_size != m_last_font_size)
```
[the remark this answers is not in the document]
amokiche-amd: Agree with copilot, I understood that impossible for us now when we use actualy integer values for font, but it can be changed in the future and this might be a protentional hard to debug problem.

src/view/src/rocprofvis_font_manager.cpp
```diff
@@ -19,43 +18,90 @@ namespace View
...
                                               14.0f, 15.0f, 16.0f, 17.0f, 18.0f,
                                               19.0f, 20.0f, 21.0f, 22.0f, 23.0f,
                                               25.0f, 27.0f, 29.0f, 31.0f, 35.0f };
 
-// Size offsets applied to the base index to produce kSmall/kMedium/kMedLarge/kLarge.
+// Offsets applied to the user-selected base index (kMedium) to produce
+// kSmall/kMedium/kMedLarge/kLarge.
 static constexpr int kSizeOffsets[FontManager::kNumSizes] = { -1, 0, 1, 2 };
 
+static int
```
amokiche-amd: What the point to make it static? It called only two times inside this class and it allow to get rid of the first argument and use the m_available_sizes directly?

src/view/src/rocprofvis_font_manager.cpp
```diff
@@ -19,43 +18,90 @@ namespace View
...
+            float delta = std::abs(available_sizes[i] - font_size);
+            if(delta < best_delta)
+            {
+                best_delta = delta;
+                best_index = i;
+            }
+        }
+    }
+
+    return best_index;
```
amokiche-amd: I understand the loop only has 20 iterations, and optimizations aren't really needed, but we're writing in C++, so I'll suggest it, and it's up to you to decide whether to keep it or not. You could avoid iterating through the entire loop and simply use the algorithm from the standard library and then simply compare adjacent indices, something like this:

static int
GetClosestFontSizeIndex(const std::vector<float>& available_sizes, float font_size)
{
    if(available_sizes.empty())
        return 0;

    auto it = std::lower_bound(available_sizes.begin(), available_sizes.end(), font_size);

    if(it == available_sizes.begin())
        return 0;
    if(it == available_sizes.end())
        return (available_sizes.size() - 1);

    auto prev = std::prev(it);
    if((font_size - *prev) <= (*it - font_size))
        return std::distance(available_sizes.begin(), prev);
    return std::distance(available_sizes.begin(), it);
}

at least it not necessary allocate and dealocate float delta 20 times inside loop

src/view/src/rocprofvis_font_manager.cpp
```diff
@@ -19,43 +18,90 @@ namespace View
...
-           std::abs(m_available_sizes[i - 1] - scaled_size))
-            best_index = i;
-    }
-    return best_index;
+int
+FontManager::ClampFontSizeIndex(int idx) const
+{
+    int min_idx = GetFontSizeIndex(MIN_USER_FONT_SIZE);
+    int max_idx = GetFontSizeIndex(MAX_USER_FONT_SIZE);
+    return std::max(min_idx, std::min(idx, max_idx));
```
amokiche-amd:
```suggestion
   return std::clamp(idx, GetFontSizeIndex(MIN_USER_FONT_SIZE), GetFontSizeIndex(MAX_USER_FONT_SIZE));
```
you can use std::clamp here

tomk-amd:
I think Derek mentioned this on the mulit-view PR that had a version of this DPI refactor included... The mini-map is tiny.  The mini map window needs to be bigger.

image

src/view/src/widgets/rocprofvis_gui_helpers.cpp
```diff
@@ -506,9 +506,10 @@ DrawInternalBuildBanner(const char* text /*= "Internal Build"*/)
     ImDrawList*   dl   = ImGui::GetForegroundDrawList();
     const ImVec2& disp = ImGui::GetIO().DisplaySize;
 
-    // Parameters
-    static constexpr float ribbon_thickness = 20.0f;
-    static constexpr float min_base_length  = 150.0f;
+    // Parameters. Scale with the font so the banner tracks ImGui's DPI font scaling.
+    const float            ui_scale         = ImGui::GetFontSize() / 15.0f;
+    const float            ribbon_thickness = 20.0f * ui_scale;
```
tomk-amd, line 568: Which is the correct constant value? 13.0 or 15.0
 
In the another function:

```
ImVec2
GetResponsiveWindowSize(ImVec2 desired_size, ImVec2 min_size, float viewport_margin)
{
    constexpr float BASE_DESIGN_FONT_SIZE = 13.0f;
    const float     scale = ImGui::GetFontSize() / BASE_DESIGN_FONT_SIZE;

    ImVec2 result(desired_size.x > 0.0f ? desired_size.x * scale : desired_size.x,
                  desired_size.y > 0.0f ? desired_size.y * scale : desired_size.y);
```

tomk-amd (approved):

PR #896, opened 2026-06-24
Redesign timeline annotations: track-bound notes with inline editing and floating windows

dhingora-amd:
## Motivation

Timeline annotations (sticky notes) were clunky to use. Creating and editing
a note required separate "Add" and "Edit" modal popups, notes floated free in
content space rather than belonging to any track, and the timeline anchor was
hard to place — it couldn't be moved while a note was expanded and couldn't be
dragged past the visible range. This PR reworks annotations end-to-end into a
direct, inline experience and ties each note to the track it describes.

## Technical Details

- **Track-bound notes:** Annotations are now anchored to a specific track,
  with their Y position stored relative to that track's top. The bound
  `track_id` is persisted (`JSON_KEY_ANNOTATION_TRACK_ID`) and `NavigationEvent`
  carries it so "go to note" lands on the right track. The annotation list
  gains a Track column showing the bound track's name, or "Unbound" for legacy
  notes saved before this change.
- **Inline create/edit:** Removed the Add/Edit modal dialogs (and the
  `kStickyNoteEdited` round-trip). A new note opens expanded and focused,
  commits on the first keystroke, and auto-discards if left empty. The
  expanded note's title is click-to-edit and the body is inline-editable, with
  trash/close buttons in the window.
- **Floating expanded window:** The expanded note is now a movable floating
  window that stays where the user drags it (position tracked live, not
  persisted).
- **Draggable anchor + edge auto-scroll:** The timeline anchor can be dragged
  whether the note is minimized or expanded. It is clamped to the visible
  track viewport, and the timeline pans/scrolls when the anchor nears a
  viewport edge so it can be placed beyond the currently visible range
  (rendering stays continuous during the drag for smooth motion).
- **Time guide line:** A hovered/dragged note draws a time indicator line
  spanning the full visible track area.
- **Restyle:** Reworked the sticky-note color palette (dark and light themes)
  to a warmer amber look with better contrast.
- **Cleanup:** Removed dead annotation code (modal/edit paths, unused members,
  empty destructor, stale includes), added an `INVALID_STICKY_ID` sentinel,
  and updated the agent docs in `.agents/AGENTS.md`.

### DIFF
--- a/src/view/src/rocprofvis_annotation_view.cpp
+++ b/src/view/src/rocprofvis_annotation_view.cpp
@@ -4,6 +4,7 @@
 #include "rocprofvis_annotation_view.h"
 #include "icons/rocprovfis_icon_defines.h"
 #include "rocprofvis_data_provider.h"
+#include "rocprofvis_event_manager.h"
 #include "rocprofvis_events.h"
 #include "rocprofvis_settings_manager.h"
 #include "rocprofvis_utils.h"
--- a/src/view/src/rocprofvis_annotation_view.cpp
+++ b/src/view/src/rocprofvis_annotation_view.cpp
@@ -117,9 +117,11 @@ AnnotationView::Render()
             ImGui::PopID();
 
             // Track column: the bound track's display name, or a placeholder
-            // for unbound legacy notes.
+            // for unbound legacy notes. Match the Text column's vertical offset
+            // so ElidedText lines up with the other columns.
             ImGui::TableNextColumn();
-            ImGui::AlignTextToFramePadding();
+            ImGui::SetCursorPosY(ImGui::GetCursorPosY() +
+                                 ImGui::GetStyle().FramePadding.y);
             const std::string track_name =
                 m_data_provider.DataModel().BuildTrackName(note.GetTrackId());
             if(track_name.empty())
@@ -140,12 +142,16 @@ AnnotationView::Render()
                                         true);
             ImGui::TextUnformatted(time_label.c_str());
 
-            // Visibility column
+            // Visibility column. When the bound track is hidden the note is
+            // force-hidden on the timeline, so disable (grey out) the toggle.
             ImGui::TableNextColumn();
-            bool        visible     = note.IsVisible();
-            std::string checkbox_id = "##visible_" + std::to_string(note.GetID());
+            bool        visible       = note.IsVisible();
+            const bool  track_hidden  = note.IsTrackHidden();
+            std::string checkbox_id   = "##visible_" + std::to_string(note.GetID());
+            if(track_hidden) ImGui::BeginDisabled();
             ImGui::Checkbox(checkbox_id.c_str(), &visible);
-            if(visible != note.IsVisible()) note.SetVisibility(visible);
+            if(track_hidden) ImGui::EndDisabled();
+            if(!track_hidden && visible != note.IsVisible()) note.SetVisibility(visible);
 
             ImGui::PopID();
         }
--- a/src/view/src/rocprofvis_annotation_view.cpp
+++ b/src/view/src/rocprofvis_annotation_view.cpp
@@ -3,6 +3,7 @@
 
 #include "rocprofvis_annotation_view.h"
 #include "icons/rocprovfis_icon_defines.h"
+#include "rocprofvis_data_provider.h"
 #include "rocprofvis_events.h"
 #include "rocprofvis_settings_manager.h"
 #include "rocprofvis_utils.h"
@@ -38,13 +39,15 @@ AnnotationView::Render()
                              0.5f);
         ImGui::TextDisabled("No annotations.");
     }
-    else if(ImGui::BeginTable("StickyNotesTable", 4,
+    else if(ImGui::BeginTable("StickyNotesTable", 5,
                               ImGuiTableFlags_BordersOuter | ImGuiTableFlags_BordersV |
                                   ImGuiTableFlags_RowBg | ImGuiTableFlags_Resizable |
                                   ImGuiTableFlags_SizingStretchProp))
     {
         ImGui::TableSetupColumn("Title");
         ImGui::TableSetupColumn("Text");
+        ImGui::TableSetupColumn("Track", ImGuiTableColumnFlags_WidthFixed,
+                                ImGui::GetFontSize() * 10.0f);
         ImGui::TableSetupColumn("Time", ImGuiTableColumnFlags_WidthFixed,
                                 ImGui::GetFontSize() * 12.0f);
         ImGui::TableSetupColumn("Visible", ImGuiTableColumnFlags_WidthFixed,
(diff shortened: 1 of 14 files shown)


tomk-amd:
When the track that the note is attached to is hidden the note lands on a different track.  And it still shows the old (hidden) track name as its "parent", ex:

image

Once it is clicked it updates to lock onto the new track.  I am not sure if this is correct? When the track is hidden, I think the annotation should hide along with the track?  Thoughts?

src/view/src/rocprofvis_annotations.cpp
```diff
@@ -74,6 +79,8 @@ AnnotationsManagerProjectSettings::ToJson()
         sticky_json[JSON_KEY_ANNOTATION_TEXT]             = notes[i].GetText();
         sticky_json[JSON_KEY_ANNOTATION_TITLE]            = notes[i].GetTitle();
         sticky_json[JSON_KEY_ANNOTATION_ID]               = notes[i].GetID();
+        sticky_json[JSON_KEY_ANNOTATION_TRACK_ID] =
+            static_cast<double>(notes[i].GetTrackId());
```
drchen-amd: Double for track id?

src/view/src/rocprofvis_annotation_view.cpp
```diff
@@ -111,6 +116,23 @@ AnnotationView::Render()
...
+            const std::string track_name =
+                m_data_provider.DataModel().BuildTrackName(note.GetTrackId());
+            if(track_name.empty())
+            {
+                ImGui::TextDisabled("Unbound");
+            }
+            else
+            {
+                ImGui::PushID("note_track");
+                ElidedText(track_name.c_str(), ImGui::GetContentRegionAvail().x);
```
drchen-amd, line 137: This column has different vertical alignment vs others.

Image

src/view/src/rocprofvis_stickynote.cpp
```diff
@@ -99,334 +215,449 @@ StickyNote::IsVisible() const
...
         ImGui::PushStyleColor(ImGuiCol_Text, muted_text_color);
-        if(ImGui::Button((std::string(ICON_EDIT) + "##" +
-                          std::to_string(reinterpret_cast<uintptr_t>(this)))
-                             .c_str(),
-                         ImVec2(action_btn_size, action_btn_size)))
+
+        ImGui::SetCursorPos(ImVec2(
+            win_size.x - action_btn_size * 2.0f - kHeaderButtonGap - kNoteMargin,
+            action_btn_y));
+        if(ImGui::Button(
```
drchen-amd: These buttons  become off center from frame the larger the font. Can we center or just go frameless?

Image

drchen-amd (commented):
We can remove `StickyNoteEvent` in rocprofvis_events.cpp. This is no longer used in current implementation.

drchen-amd:
> When the track that the note is attached to is hidden the note lands on a different track. And it still shows the old (hidden) track name as its "parent", ex:
> 
> image
> Once it is clicked it updates to lock onto the new track. I am not sure if this is correct? When the track is hidden, I think the annotation should hide along with the track? Thoughts?

Maybe when a track with a note is hidden, the square indicator hides with it, and in the annotation table, the visible checkbox is greyed out.

src/view/src/rocprofvis_stickynote.cpp
```diff
@@ -437,102 +669,46 @@ StickyNote::HandleDrag(const ImVec2&                       window_position,
...
+            max_y = layout.view_max_y - drag_h;
+        }
+        new_y = std::clamp(new_y, min_y, std::max(min_y, max_y));
+
+        m_time_ns = conversion_manager->PixelToTime(new_x);
+        // Re-anchor to the track under the cursor, storing a track-relative
+        // offset so the note follows reorder / collapse / resize.
+        uint64_t track_id    = INVALID_TRACK_ID;
+        float    track_top_y = 0.0f;
+        if(layout.track_at && layout.track_at(new_y, track_id, track_top_y))
```
drchen-amd: I suggest for this dragging case; instead of constantly anchoring to track while dragging, we just do it once when dragging stops.

I expect this will kill the annotations table live updating the track column as you drag, but considering each `track_at` loops all graphs, is it not cheap so I don't think it worth doing. I think it would be fine while dragging, the track column turns grey and keeps displaying the value from before drag started. Or maybe while dragging it just shows "-" or something.

drchen-amd (commented):
Inputs from overlapping windows like preferences dialog or about dialog fall through to sticky anchor.

src/view/src/rocprofvis_stickynote.cpp
```diff
@@ -653,12 +660,14 @@ StickyNote::HandleDrag(const ImVec2&                       window_position,
     }
 
     if((dragged_id == INVALID_STICKY_ID || dragged_id == m_id) && !m_dragging &&
-       !over_expanded_window && ImGui::IsMouseHoveringRect(icon_pos, drag_max) &&
+       !over_expanded_window && !blocked_by_popup &&
+       ImGui::IsMouseHoveringRect(icon_pos, drag_max) &&
```
drchen-amd: These `IsMouseHoveringRect` maybe causing the input problems.

You can try `ImGui::IsWindowHovered` inside your anchor window. In the past we've used this to deal with the click through problems so you can search for this in the code for flags and examples.

I think setting your `m_dragging` to anchor window hovered && mouse down could be enough, and you wouldn't need the  `!over_expanded_window && !blocked_by_popup`

drchen-amd (commented):
Input still falls through to anchor for windows like minimap, logviewer, summary.

drchen-amd:
If you increase the height of track, place a note in the revealed region, then shrink the track, the anchor appears on top of a different track.

[Screencast from 06-29-2026 03:58:22 PM.webm](https://github.com/user-attachments/assets/f77d1c73-e174-41b4-bcc8-b4a00d3f0b30)

I think the anchor should be allowed to be cutoff or not rendered depending on track height.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -250,6 +250,19 @@ TimelineView::BuildTrackLayout()
...
+        {
+            if(graph.chart && graph.chart->GetID() == track_id)
+            {
+                out_height = graph.chart->GetTrackHeight();
+                return true;
+            }
+        }
+        return false;
+    };
+
```
drchen-amd: I see this is being used like `layout.height_of(sticky.m_track_id, out).`

In your `EnsureBounds` where you update sticky's `m_track_id`, what if instead of storing track id, we store ptr to `TrackItem`? Then instead of having this search every frame, you read track height from ptr.

drchen-amd (commented):
There is a deadzone below anchor where the track height resize will not respond. It seems to be the shape/size of the expanded note:

https://github.com/user-attachments/assets/1628bb5d-4e9b-4476-aa81-1294db9030dc

The clipping at the boundaries feels a too aggressive? My anchor is ~90% inside the lower track, but after letting go it settles in the higher track.

https://github.com/user-attachments/assets/8bd000cc-fbfa-463c-a79a-d29c69c43670

tomk-amd (approved):

PR #903, opened 2026-06-25
Add right-click copy to Track Details tables

dhingora-amd:
## Motivation

The Track Details tab had no way to copy its values. The Event Details
tab already supports right-click copy on its tables, so Track Details
was inconsistent and users couldn't easily grab a track's info or its
computed stats. This brings Track Details to parity.

## Technical Details

Track Details rows now support a right-click context menu with "Copy Row
Data" / "Copy Cell Data", each raising the standard copy notification —
matching the Event Details tab.

- Applies to both the topology/track info tables and the computed stats
  rows (queue utilization, and counter min/max/mean/stddev).
- Reuses the shared right-click hitbox and copy-menu scaffolding in
  `gui_helpers` (`RenderRowHitbox`, `PositionCell`,
  `CaptureCellRightClick`, `BeginCellContextMenu`,
  `AddCopyRowCellMenuItems`) added in #877, so no copy/menu/notification
  logic is duplicated.
- Each table is given a unique id so its context-menu popup is distinct,
  and the stats rows share a single `CellMenuTarget` via an offset row
  index. Existing cell elide/expand behavior is preserved.

### DIFF
--- a/src/view/src/rocprofvis_track_details.cpp
+++ b/src/view/src/rocprofvis_track_details.cpp
@@ -12,6 +12,10 @@
 #include "widgets/rocprofvis_gui_helpers.h"
 #include "widgets/rocprofvis_widget.h"
 
+#include <cstdio>
+#include <string>
+#include <vector>
+
 namespace RocProfVis
 {
 namespace View
@@ -166,13 +170,15 @@ TrackDetails::Render()
                             {
                                 ImGui::Text("Node: %s",
                                             detail.parents.node->info->host_name.c_str());
-                                RenderTable(detail.parents.node->info_table);
+                                RenderTable(detail.parents.node->info_table,
+                                            "##td_node_table");
                             }
                             if(detail.parents.process)
                             {
                                 ImGui::Text("Process: %s",
                                             detail.parents.process->header.c_str());
-                                RenderTable(detail.parents.process->info_table);
+                                RenderTable(detail.parents.process->info_table,
+                                            "##td_process_table");
                             }
                         }
                         ImGui::EndChild();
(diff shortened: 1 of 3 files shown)


src/view/src/rocprofvis_track_details.cpp
```diff
@@ -392,46 +420,66 @@ TrackDetails::RenderTable(InfoTable& table, const AnalysisTrackStatistics* stats
...
-                                                 AnalysisTrackStatistics::kReady);
-                            ImGui::Text("%.1f", stats->stats[i].value);
-                            ImGui::EndDisabled();
-                        }
-                        break;
+                        const InfoTable::Cell& cell = table.cells[m_cell_menu.row][c];
+                        row_cells.push_back(cell.needs_format ? cell.formatted
+                                                              : cell.data);
                     }
+                    AddCopyRowCellMenuItems(row_cells.data(), cols, m_cell_menu.column);
```
drchen-amd, line 482: For the copy row case, this AddCopyRowCellMenuItems and whatever InfiniteScrollTable uses should use the same seperator.  

Currently InfiniteScrollTable are comma seperated, but AddCopyRowCellMenuItems is space sepearted.

src/view/src/rocprofvis_track_details.cpp
```diff
@@ -392,46 +420,66 @@ TrackDetails::RenderTable(InfoTable& table, const AnalysisTrackStatistics* stats
...
+                    PositionCell(1);
+                    ImGui::BeginDisabled(!stats_ready);
+                    ImGui::Text("%.1f", stats->stats[i].value);
+                    ImGui::EndDisabled();
+                    CaptureCellRightClick(1, stat_row, m_cell_menu, open_menu);
+                    ImGui::PopID();
+                }
+            }
+
+            const std::string ctx_menu_id = std::string(table_id) + "_ctx";
```
drchen-amd: No need to build this string id. You can just define a constant id because there will only be one active at a time. It looks like this is the only user of `table_id` parameter, so that could be removed.

drchen-amd (approved):

PR #904, opened 2026-06-25
Compare Traces

Sajeeth-Wimalasuriyan:
image
image

### DIFF
--- a/src/model/src/database/rocprofvis_db.cpp
+++ b/src/model/src/database/rocprofvis_db.cpp
@@ -554,7 +554,7 @@ void Database::CreateTracksOrderRanking() {
         partitions[t->load_id].push_back(t);
     }
 
-    uint32_t partition_id = 0;
+    uint32_t order_base = 0;
 
     for (auto& [load_set, partition_tracks] : partitions) {
 
@@ -585,13 +585,17 @@ void Database::CreateTracksOrderRanking() {
         for (size_t i = 0; i < max_len; ++i) {
             for (auto& [db, vec] : groups) {
                 if (i < vec.size()) {
-                    vec[i]->order_id = partition_id * 100 + local_order;
+                    vec[i]->order_id = order_base + local_order;
                     ++local_order;
                 }
             }
         }
 
-        ++partition_id;
+        // Advance the base past every track assigned in this partition so
+        // order_ids stay globally unique regardless of how many tracks a
+        // partition holds (a fixed stride per partition would collide once a
+        // partition exceeds that stride).
+        order_base += local_order;
     }
 
 }
--- a/src/model/src/database/rocprofvis_db.cpp
+++ b/src/model/src/database/rocprofvis_db.cpp
@@ -563,7 +563,8 @@ void Database::CreateTracksOrderRanking() {
 
         for (auto* t : partition_tracks) {
             DbInstance* db_instance = (DbInstance*)t->track_indentifiers.db_instance;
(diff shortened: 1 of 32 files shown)


amokiche-amd:
Not a comment, but a small feature request. I think when someone is working with a trace and clicks the compare button, it would be nice to reduce the number of clicks and automatically set the "first" field to the currently open trace(or the last open one from the list). Keep the browse button to allow editing; if no traces are open, keep the current state with two empty fields.

src/model/src/database/rocprofvis_db.cpp
```diff
@@ -542,6 +542,60 @@ Database::UpdateQueryForTrack(  rocprofvis_dm_track_params_it it,
...
+        for (const auto& [_, vec] : groups) {
+            max_len = std::max(max_len, vec.size());
+        }
+
+        uint32_t local_order = 0;
+
+        for (size_t i = 0; i < max_len; ++i) {
+            for (auto& [db, vec] : groups) {
+                if (i < vec.size()) {
+                    vec[i]->order_id = partition_id * 100 + local_order;
```
amokiche-amd: Indexes will be invalid if the number of tracks in one of the files is more than 100. I don't know how realistic this is, but it is a potential problem.

src/model/src/database/rocprofvis_db.cpp
amokiche-amd, line 481: What the point of this function?

src/model/src/database/rocprofvis_db_rocprof.cpp
```diff
@@ -1207,6 +1207,10 @@ rocprofvis_dm_result_t  RocprofDatabase::ReadTraceMetadata(Future* future)
                 t.join();
             load_id++;
         }
+
+        // Crerate track rankings based on load_id abd bd instance
```
amokiche-amd:
```suggestion
        // Create track rankings based on load_id abd bd instance
```
typo

src/model/src/database/rocprofvis_db_rocpd.cpp
```diff
@@ -301,6 +301,9 @@ rocprofvis_dm_result_t  RocpdDatabase::ReadTraceMetadata(Future* future)
             },
                         &CallBackAddTrack, &CallBackLoadTrack)) break;
 
+        // Crerate track rankings based on load_id and db instance
```
amokiche-amd:
```suggestion
        // Create track rankings based on load_id and db instance
```
typo

src/model/src/database/rocprofvis_db.cpp
```diff
@@ -542,6 +542,60 @@ Database::UpdateQueryForTrack(  rocprofvis_dm_track_params_it it,
...
+    uint32_t partition_id = 0;
+
+    for (auto& [load_set, partition_tracks] : partitions) {
+
+
+        std::map<uint32_t, std::vector<Track*>> groups;
+
+        for (auto* t : partition_tracks) {
+            DbInstance* db_instance = (DbInstance*)t->track_indentifiers.db_instance;
+            uint32_t file_index = db_instance ? db_instance->FileIndex() : 0;
```
amokiche-amd, line 566: I don't know how db_instance can become zero, but since it's being checked, it's theoretically possible (at least in the future). This means that if db_instance is zero for both files, we'll use file_index = 0 for both files and combine all the tracks into one groop.

tomk-amd (approved):

PR #906, opened 2026-06-26
Block app launch on Intel Macs with a native alert

Sajeeth-Wimalasuriyan:
Display error message when app is opened on x86 based mac systems. We do not officially support this.


tomk-amd:
Not sure about doing this.. people can build themselves for x86 based macs and everything should still work

PR #907, opened 2026-06-26
Add right-click copy to Kernel Selection Table

dhingora-amd:
## Motivation

The Kernel Selection Table (a.k.a. Kernel Pivot Table) in the Compute view did not support copying data, unlike the other tables in the app (Event Details, Track Details, etc.). Users had no quick way to pull a kernel's metric values out of the table for sharing, notes, or comparison. This PR brings the Kernel Selection Table in line with the rest of the app's copy behavior.

## Technical Details

- Added right-click **Copy Row Data** / **Copy Cell Data** context menu support to the Kernel Selection Table, reusing the shared copy helpers (`CellMenuTarget`, `BeginCellContextMenu` / `AddCopyRowCellMenuItems` / `EndCellContextMenu`) so behavior and copy notifications match the other tables.
- The column under the cursor is resolved geometrically from each cell's rectangle bounds rather than `TableGetHoveredColumn()`, which is unreliable with frozen columns and a `SpanAllColumns` selectable. This ensures the correct column's value is copied (previously, copying a cell like Duration could copy the Name column).
- Row and cell copies exclude hidden/disabled columns (e.g. the internal ID column), with the cell index remapped to the visible column set. Visible-column info is captured before the popup opens to avoid reading table state in an invalid context.
- Empty cells copy as `N/A`; copied row values are comma-separated for consistency with the other tables.
- Existing left-click kernel selection behavior is unchanged.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
+++ b/src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
@@ -692,18 +692,6 @@ KernelMetricTable::Render()
                     ImGui::EndDisabled();
                 }
 
-                // Collect visible columns here; TableGetColumnFlags is invalid once
-                // the popup window is open.
-                std::vector<int> visible_cols;
-                visible_cols.reserve(column_count);
-                for(int c = 0; c < column_count; c++)
-                {
-                    if(ImGui::TableGetColumnFlags(c) & ImGuiTableColumnFlags_IsEnabled)
-                    {
-                        visible_cols.push_back(c);
-                    }
-                }
-
                 if(open_menu)
                 {
                     ImGui::OpenPopup(CELL_CONTEXT_MENU_ID);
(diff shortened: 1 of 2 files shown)


src/view/src/compute/rocprofvis_compute_kernel_metric_table.cpp
```diff
@@ -678,6 +692,52 @@ KernelMetricTable::Render()
...
+                // the popup window is open.
+                std::vector<int> visible_cols;
+                visible_cols.reserve(column_count);
+                for(int c = 0; c < column_count; c++)
+                {
+                    if(ImGui::TableGetColumnFlags(c) & ImGuiTableColumnFlags_IsEnabled)
+                    {
+                        visible_cols.push_back(c);
+                    }
+                }
```
drchen-amd: If you use the header names instead of flags, you could move this inside the context menu block. Your loop would become:

```
const std::vector<std::string>& menu_row = data[m_cell_menu.row];
std::vector<std::string>        row_cells;
// Remap the data-column index to its visible-list position.
int cell_index = 0;
for(size_t col = 0; col < header.size(); col++)
{
    if(!header[col].empty() && header[col][0] != '_')
    {
        if(col == m_cell_menu.column)
        {
            cell_index = static_cast<int>(row_cells.size());
        }
        row_cells.push_back(menu_row[col].empty() ? "N/A"
                                            : menu_row[col]);
    }
}
```

drchen-amd (approved):

PR #916, opened 2026-06-30
Add usage tooltip to Compute comparison threshold input

dhingora-amd:
## Motivation

The Compute comparison threshold input is a drag-to-edit control, but
nothing on screen tells the user they can drag it or double-click to
type a value, so the interaction isn't discoverable.

## Technical Details

- Added a tooltip to the Compute comparison threshold box reading
  "Drag to adjust or double click to edit." alongside the existing
  threshold description.
- Made hover tooltips render reliably when the app is idle so the new
  hint (and other tooltips) actually show up.

### DIFF
--- a/src/view/src/rocprofvis_appwindow.cpp
+++ b/src/view/src/rocprofvis_appwindow.cpp
@@ -597,13 +597,6 @@ AppWindow::WantsContinuousRender()
         return true;
     }
 
-    // Keep rendering while an item is hovered so the tooltip hover-delay timer
-    // can elapse instead of the idle loop sleeping before tooltips appear.
-    if(ImGui::IsAnyItemHovered())
-    {
-        return true;
-    }
-
 #ifdef ROCPROFVIS_DEVELOPER_MODE
     if(m_test_data_provider.GetState() == ProviderState::kLoading ||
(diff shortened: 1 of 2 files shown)


src/view/src/rocprofvis_appwindow.cpp
```diff
@@ -597,6 +597,13 @@ AppWindow::WantsContinuousRender()
         return true;
     }
 
+    // Keep rendering while an item is hovered so the tooltip hover-delay timer
+    // can elapse instead of the idle loop sleeping before tooltips appear.
+    if(ImGui::IsAnyItemHovered())
+    {
+        return true;
+    }
+
```
tomk-amd: This keeps the render loop alive if any "interactive" element is hovered, including table cells, scroll bars, menu items etc..

Is this still necessary with the wait for poll with wait change that was recently pushed: https://github.com/ROCm/roc-optiq/pull/913 (this branch does not have it yet)?

dhingora-amd: Maybe not then, I just added it since the tooltips wasn't showing reliably... but if the #913 fixes it, it might not be necessary... I'll go ahead and remove it.

tomk-amd (approved):

PR #919, opened 2026-07-03
[PoC]: Add dedicated per-GPU KFD tracks

sputhala-amd:
Render KFD events (page fault, queue, unmap, migrate) as dedicated per-GPU tracks under each GPU node instead of on the CPU main-thread track. Wires the KFD query builders into track discovery, topology routing (TopologyNodeKfd), track naming/category, controller track creation and slice fetch, and view track-name/label handling.

## Submission Checklist

### DIFF
--- a/src/model/src/database/rocprofvis_db_query_factory.cpp
+++ b/src/model/src/database/rocprofvis_db_query_factory.cpp
@@ -31,6 +31,9 @@ namespace DataModel
                 
                 { Builder::From("rocpd_region", "R"),
                 Builder::InnerJoin("rocpd_track", "T", "T.id = R.track_id"),
+                // Exclude KFD events; they are rendered on dedicated per-GPU KFD tracks.
+                Builder::InnerJoin("rocpd_event", "KE", "KE.id = R.event_id"),
+                Builder::InnerJoin("rocpd_string", "KCAT", "KCAT.id = KE.category_id AND KCAT.string NOT LIKE 'rocm_kfd_%'"),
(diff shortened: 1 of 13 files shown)


sputhala-amd:
This is currently only a draft or POC created for discussion purposes.

src/model/src/database/rocprofvis_db_query_factory.cpp
```diff
@@ -31,6 +31,9 @@ namespace DataModel
                 
                 { Builder::From("rocpd_region", "R"),
                 Builder::InnerJoin("rocpd_track", "T", "T.id = R.track_id"),
+                // Exclude KFD events; they are rendered on dedicated per-GPU KFD tracks.
+                Builder::InnerJoin("rocpd_event", "KE", "KE.id = R.event_id"),
+                Builder::InnerJoin("rocpd_string", "KCAT", "KCAT.id = KE.category_id AND KCAT.string NOT LIKE 'rocm_kfd_%'"),
```
drchen-amd, line 36: I suppose this is a downside with having this data in region table, as opposed to being in its own table. Both standard region queries, and these kfd queries must do this extra filtering work to isolate themselves.

drchen-amd (commented):
Could you please share sample DB with this data?

sputhala-amd:
> Could you please share sample DB with this data?

Thanks @drchen-amd, I haven't yet tested this solution.
The purpose of this PR is only for an initial discussion around how KFD data can be represented on separate tracks and what is the drawback of the current DB schema, as you correctly identified.

PR #920, opened 2026-07-06
View: Color-code tracks and sidebar by node

dhingora-amd:
## Motivation

On multi-node traces it's hard to tell which node a given track belongs to.
Track meta labels don't surface node identity clearly, and there's no visual
link between the sidebar topology tree and the tracks on the timeline. This
PR adds subtle, consistent node color-coding so users can immediately see how
tracks group by node and correlate them with the sidebar.

## Technical Details

- **Track pill (`rocprofvis_track_item`):** Added `SetNodeColor`, which pulls
  the track's node from the topology model and builds a pill in the track meta
  area showing the node's display index. The pill is accented with a color-wheel
  slot keyed by that index and carries a tooltip of the form `Node <n>: <name>`.
- **Sidebar (`rocprofvis_sidebar` / `rocprofvis_tree_node`):** Node rows now
  draw a matching color swatch, so the topology tree and the timeline agree on
  which node is which. `TreeNode` gained `show_color_swatch` / `color_index`.
- **Topology model (`rocprofvis_topology_model`):** Added `GetNodeDisplayIndex`,
  returning a stable 1-based index per node used as the color-wheel key across
  the pill and swatch.
- **Preferences (`rocprofvis_settings_manager` / `rocprofvis_settings_panel`):**
  Added a "Color-code tracks by node" display setting (defaults on), toggleable
  from Edit → Preferences and persisted to config. The pill and swatch are
  gated on this preference at render time.
- **Colors** come from the shared theme color wheel, so they follow the active
  theme (dark/light) and stay consistent across the pill, swatch, and any
  future node cues.

### DIFF
--- a/src/view/src/rocprofvis_sidebar.cpp
+++ b/src/view/src/rocprofvis_sidebar.cpp
@@ -9,6 +9,8 @@
 #include "rocprofvis_settings_manager.h"
 #include "rocprofvis_timeline_selection.h"
 
+#include <cmath>
+#include <string>
 #include <unordered_set>
 
 namespace RocProfVis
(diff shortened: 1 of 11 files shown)


Sajeeth-Wimalasuriyan:
image
High res monitor has overlapping issue

drchen-amd (commented):
Looks good. 
Just want to confirm that we want this option to affect single node files?

src/view/src/rocprofvis_sidebar.cpp
```diff
@@ -369,27 +371,37 @@ SideBar::RenderBranchNode(const TreeNode& node, const TreeNode* state_node,
...
+            // the swatch into that gap so the row reads: arrow, swatch, label.
+            const float radius  = ImGui::GetTextLineHeight() * NODE_SWATCH_RADIUS_SCALE;
+            const float gap     = ImGui::GetStyle().ItemInnerSpacing.x;
+            const float space_w = ImGui::CalcTextSize(" ").x;
+            const int   pad     = space_w > 0.0f
+                                      ? static_cast<int>(std::ceil((radius * 2.0f + gap) / space_w))
+                                      : 0;
+
+            const ImVec2 node_pos = ImGui::GetCursorScreenPos();
+            open = ImGui::TreeNodeEx((std::string(pad, ' ') + node.label).c_str(),
```
drchen-amd: Side effect of using different IDs for decorator on/off: 
When you make changes to collapsed state of side bar items, then toggle the decorator option, all your collapsed states reset. 

I suggest keeping ID constant and applying padding via spacing instead of string.  This whole function is already ID guarded so you can use empty ID.

```
open = ImGui::TreeNodeEx("", HEADER_FLAGS);
ImGui::SameLine(0.0f, draw_swatch ? whatever_paddding : 0.0f);
ImGui::TextUnformatted(node.label.c_str());
```

drchen-amd: On the other hand if you apply my suggestion, then I think only the arrow will respond to inputs...

src/view/src/rocprofvis_sidebar.cpp
```diff
@@ -375,15 +369,59 @@ SideBar::RenderBranchNode(const TreeNode& node, const TreeNode* state_node,
...
+
     bool open = true;
     if(node.collapsable)
     {
-        open = ImGui::TreeNodeEx(node.label.c_str(), HEADER_FLAGS);
+        if(draw_swatch)
+        {
+            const ImGuiTreeNodeFlags flags =
+                (HEADER_FLAGS & ~ImGuiTreeNodeFlags_SpanLabelWidth) |
+                ImGuiTreeNodeFlags_SpanAvailWidth;
```
drchen-amd: Problem with this is that only the node level + decorator on will have this span. All others will not.

Tom had an idea yesterday where instead of drawing this circle swatch, we just change the color of the header's triangle. Lets try that? I suspect you might have to overlap a drawlist triangle and rotate it based on open/close because I don't think there is an api/property for the color of triangle.

drchen-amd: Another idea if the triangle drawing code is too ugly:
If we manage the collapsed state ourselves, the ID could be whatever you want and your swatch code could stay. TreeNode class could track it via bool. These headers can be programatically collapsed/opened via `ImGui::SetNextItemOpen`

drchen-amd (approved):

PR #924, opened 2026-07-07
Improve line track Y-axis scale

Sajeeth-Wimalasuriyan:
image
Added y axis labels that get generated or removed as graph is adjusted in size.


src/view/src/rocprofvis_line_track_item.cpp
```diff
@@ -145,6 +153,23 @@ LineTrackItem::BoxPlotRender(float graph_width)
     ImU32 outline_color     = alt_fill_color;
     ImU32 accent            = m_settings.GetColor(Colors::kAccent);
 
+    // Grid lines behind the data, matching the meta-area ticks.
+    if(m_track_height >= Y_AXIS_LABEL_MIN_TRACK_HEIGHT)
+    {
+        std::vector<double> grid_ticks;
+        GenerateYAxisTicks(content_size.y, grid_ticks);
+        const ImU32 grid_color =
+            ApplyAlpha(m_settings.GetColor(Colors::kGridColor), Y_AXIS_GRID_LINE_ALPHA);
```
tomk-amd, line 175: `grid_ticks` - make this a member `m_grid_ticks` - so it doesn't have to be re-created each frame...

Then call  GenerateYAxisTicks(content_size.y, grid_ticks); only if track size changed..

tomk-amd: Also however grid_color is calculated, maybe the alpha? isn't quite right..

The horizontal lines are brighter than the vertical grid lines.. I think they should be the same..

Image

tomk-amd:
Also, a nitpik: 

The inner (generated) numbers on the axis are mis-aligned compared to the min / max in the horizonal direction, they touch the divider line:

image

tomk-amd (approved):

src/view/src/rocprofvis_line_track_item.cpp
```diff
@@ -374,21 +398,126 @@
...
+LineTrackItem::UpdateYAxisTicks()
+{
+    const float  plot_height = m_track_content_height - m_vertical_padding * 2.0f;
+    const double min_v       = m_min_y.Value();
+    const double max_v       = m_max_y.Value();
+    const float  line_h      = ImGui::GetTextLineHeight();
+
+    // Ticks only depend on the plot height, the Y range, and the text height, so
+    // reuse the cached vector unless one of those has changed since last frame.
+    if(plot_height == m_cached_ticks_height && min_v == m_cached_ticks_min &&
```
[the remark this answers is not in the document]
Sajeeth-Wimalasuriyan: Addressed

PR #925, opened 2026-07-07
Add roofline line thickness preference

dhingora-amd:
## Motivation

The roofline chart's ceiling lines are drawn at ImPlot's default 1px weight, which is too thin to read comfortably on some displays. This adds a user-adjustable line thickness so the roofline can be made more legible.

## Technical Details

- Added a `roofline_line_thickness` field to `DisplaySettings` (serialized/deserialized with the display settings, clamped on load).
- Added a **Line thickness** slider under **Preferences → Display → Roofline** (1.0–6.0px), gated by `COMPUTE_UI_SUPPORT`.
- `Roofline::Render()` now uses the setting for the ceiling line weight instead of the hardcoded default. Default of 1.0 preserves the current appearance, including the hover emphasis.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_roofline.cpp
+++ b/src/view/src/compute/rocprofvis_compute_roofline.cpp
@@ -787,7 +787,7 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
         ImGui::EndGroup();
         float header_height = ImGui::GetItemRectSize().y + 2 * style.WindowPadding.y;
         float footer_height =
-            (m_kernel_mode == AllKernels ? 4 : 3) * ImGui::GetFrameHeightWithSpacing() +
+            (m_kernel_mode == AllKernels ? 6 : 5) * ImGui::GetFrameHeightWithSpacing() +
             2 * style.WindowPadding.y;
         ImGui::SetNextWindowSizeConstraints(
             ImVec2(menus_content_width, 0),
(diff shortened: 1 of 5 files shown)


drchen-amd:
We have a bunch of other roofline preferences accessed from the roofline's own gear menu, so I think this option being in the settings panel by itself is a bit odd. Although the existing options are not global and also not saved on app close.

(I'm not saying to change the code, just my observation.)

src/view/src/compute/rocprofvis_compute_roofline.cpp
drchen-amd: This is the reserved line height of the footer area that you're adding to. You added 2 lines so this needs to be  (m_kernel_mode == AllKernels ? 6 : 5).

Otherwise, you will get overlapped scrollbars.

src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -907,6 +912,17 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
         if(m_menus_mode == Options)
         {
             ImGui::SeparatorText("Options");
+            ImGui::PushID("line_thickness");
```
drchen-amd, line 925: I think visually would look better below the kernel scale option 

The kernel scale takes up one line, while this and the position control both take two lines.

drchen-amd (approved):

PR #930, opened 2026-07-09
Annotation UX enhancements: lock, go-to-anchor, cross-highlight, wrapping, and placement fixes

dhingora-amd:
## Motivation

The redesigned timeline annotations are floating, detached notes anchored to
tracks. While powerful, a few rough edges made them awkward to use:

- New annotations created from the right-click menu spawned wherever the
  cursor landed on the menu item, not where the user actually right-clicked.
- Once a note was positioned, it was easy to accidentally drag its anchor.
- Because notes float independently, scrolling could move an anchor out of
  sight with no quick way back to it from the note itself.
- Long note text ran off the edge instead of wrapping, and there was no way
  to preview a minimized note's contents without opening it.
- The note's left accent stripe rendered an unclean nub past the rounded
  corners.

This PR polishes the annotation experience to make notes easier to place,
protect, read, and navigate.

## Technical Details

- Right-click "Add Annotation" now anchors at the captured right-click
  position (m_context_menu_pos) instead of the live cursor.
- Added a per-note lock toggle (padlock icon, persisted via is_locked) that
  freezes the anchor against accidental drags.
- Added a "Go to Anchor" header button (ICON_ARROW_FORWARD, matching the
  app's existing navigate convention). The note signals intent via
  WantsNavigate() and TimelineView dispatches the same NavigationEvent used
  by the Annotations tab.
- Hovering an anchor previews the note's title/body in a styled tooltip
  (title elided to one line, body capped at 200 chars).
- Enabled ImGuiInputTextFlags_WordWrap on the note body so text wraps while
  editing and reflows on resize; unified title elision into a single
  ElideWithEllipsis helper using a plain "..." instead of "[...]".
- Cross-highlight: hovering/interacting with a note or its anchor outlines
  the other with a ring. The anchor ring is drawn on the timeline layer so
  an overlapping window occludes it; the note ring hugs the outside edge.
- Removed the left accent stripe (and its now-unused constants) that caused
  an unclean corner artifact.

### DIFF
--- a/src/view/src/rocprofvis_stickynote.cpp
+++ b/src/view/src/rocprofvis_stickynote.cpp
@@ -510,17 +510,19 @@ StickyNote::RenderExpandedWindow(const ImVec2& anchor_pos)
         ImVec2 trash_icon_size = ImGui::CalcTextSize(ICON_TRASH_CAN);
         ImVec2 lock_icon_size  = ImGui::CalcTextSize(ICON_LOCKED);
         ImVec2 unlock_icon_size = ImGui::CalcTextSize(ICON_UNLOCKED);
+        ImVec2 goto_icon_size   = ImGui::CalcTextSize(ICON_ARROW_FORWARD);
         ImGui::PopFont();
         const float action_btn_size = std::max(
             {kHeaderButtonMinSize, close_icon_size.x + kHeaderButtonPadding,
              trash_icon_size.x + kHeaderButtonPadding,
              lock_icon_size.x + kHeaderButtonPadding,
-             unlock_icon_size.x + kHeaderButtonPadding});
+             unlock_icon_size.x + kHeaderButtonPadding,
+             goto_icon_size.x + kHeaderButtonPadding});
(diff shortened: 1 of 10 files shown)


tomk-amd:
image
higlight cuts through the tooltip..

src/view/src/rocprofvis_stickynote.cpp
```diff
@@ -89,12 +94,41 @@ PopPlainTextInputStyle()
     ImGui::PopStyleVar(1);
     ImGui::PopStyleColor(3);
 }
+
+// Trims text to a single line fitting max_width (and max_chars), appending "..."
+// when shortened. Uses the current font for measurement.
+std::string
+ElideWithEllipsis(const std::string& text, float max_width, size_t max_chars)
+{
```
tomk-amd: Nit:  I think this would be useful as a general helper function.  Maybe this can be moved to rocprofvis_gui_helpers?

tomk-amd (approved):

PR #931, opened 2026-07-09
View: Refactor track stats into components; show real values in Track Details

dhingora-amd:
## Motivation

Track stats were stored as pre-baked strings that mixed name, value, and units
together, so the Track Details table showed abbreviated values (e.g. `1.83B MB`)
instead of the real ones (e.g. `1830730632.2 MB`), and consumers couldn't
compose their own labels.

## Technical Details

- Refactored `AnalysisTrackStatistics::Stat` into raw components (prefix,
  units suffix, compact value, full value, real double) with label helpers
  (`CompactValue` / `FullValue` / `CompactLabel` / `FullLabel`).
- `AnalysisModel` now stores the real unrounded value and fills the components
  instead of combined strings.
- Consumers build the label they need: timeline pills stay abbreviated, while
  the Track Details table shows the full real values.

### DIFF
--- a/src/view/src/rocprofvis_flame_track_item.cpp
+++ b/src/view/src/rocprofvis_flame_track_item.cpp
@@ -182,21 +182,13 @@ FlameTrackItem::Update()
         if(m_track_statistics->state == AnalysisTrackStatistics::kReady &&
            m_track_statistics_dirty)
         {
+            const AnalysisTrackStatistics::Stat& stat =
+                m_track_statistics
+                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization];
             m_pill_analysis_queue->Activate();
-            m_pill_analysis_queue->SetLabel(
-                m_track_statistics
-                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization]
-                    .compact,
-                Pill::kCompact);
-            m_pill_analysis_queue->SetLabel(
-                m_track_statistics
-                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization]
-                    .extended,
-                Pill::kExtended);
-            m_pill_analysis_queue->SetTooltip(
-                m_track_statistics
-                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization]
-                    .full);
+            m_pill_analysis_queue->SetLabel(stat.CompactValue(), Pill::kCompact);
+            m_pill_analysis_queue->SetLabel(stat.CompactLabel(), Pill::kExtended);
+            m_pill_analysis_queue->SetTooltip(stat.FullLabel());
         }
         else if(m_track_statistics->state < AnalysisTrackStatistics::kReady)
         {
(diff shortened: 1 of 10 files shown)


src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -182,21 +182,13 @@ FlameTrackItem::Update()
...
-                    .extended,
-                Pill::kExtended);
-            m_pill_analysis_queue->SetTooltip(
-                m_track_statistics
-                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization]
-                    .full);
+                    ->stats[AnalysisTrackStatistics::Queue::kQueueUtilization];
+            m_pill_analysis_queue->Activate();
+            m_pill_analysis_queue->SetLabel(stat.CompactValue(), Pill::kCompact);
+            m_pill_analysis_queue->SetLabel(stat.CompactLabel(), Pill::kExtended);
```
amokiche-amd: This comment isn't about these changes, but rather about the implementation of this function. It looks like instead of SetLabel with a switch case inside and a parameter being passed, there should be two different functions: SetCompactLabel and SetExtendedLabel.

amokiche-amd (approved):
I'll leave the decision on the comment up to you :)

tomk-amd:
image

One strange behaviour, for the "full" value strings the decimal precision is fixed to one digit.  For small values this actually makes the "full" value less precise than the compact one.  See the std dev values: 1.69 compact and 1.7 full...

src/view/src/rocprofvis_track_item.h
```diff
@@ -63,7 +63,8 @@ class Pill
     Pill(bool shown, bool active);
     ~Pill();
 
-    void   SetLabel(const std::string& label, Sizing sizing = kCompact);
+    void   SetCompactLabel(const std::string& label);
+    void   SetExtendedLabel(const std::string& label);
     void   SetTooltip(std::string label);
```
tomk-amd, line 68: Hmm - if we get rid of a common `SetLabel()`, then I think `SetCompactLabel()` should be renamed to `SetLabel()` as this will be the default choice anyway? `SetExtendedLabel()` can be kept as is...

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -695,17 +695,17 @@ TrackItem::SetDefaultPillLabel(const TrackInfo* track_info)
...
-                pill->SetLabel("THREAD");
+                pill->SetCompactLabel("THREAD");
             }
             break;
         }
         case TrackInfo::TrackType::SampledThread:
         {
-            pill->SetLabel("SAMPLED THREAD");
+            pill->SetCompactLabel("SAMPLED THREAD");
             break;
```
tomk-amd, line 709: For example here `pill->SetLabel("THREAD")` reads better than `pill->SetCompactLabel("THREAD")`

tomk-amd (approved):

PR #933, opened 2026-07-10
View: Add "Reveal in topology" track right-click action

dhingora-amd:
## Motivation

The sidebar already lets you jump from a topology node to its track in
the timeline. This adds the reverse: right-click a timeline track and
"Reveal in topology" to find its row in the tree — useful on large
traces where the row is buried in collapsed groups and off-screen.

## Technical Details

- New "Reveal in topology" track context-menu item that emits a
  `kRevealTrackInTopology` event (reusing `ScrollToTrackEvent`).
- `SideBar` handles it by expanding only the ancestor groups on the
  path, scrolling the row into view, and pulsing it with the same
  highlight animation as the event table's "go to event".
- The ancestor path is only computed during the initial expand/scroll
  frames; afterward the pulse target is matched by track id in the
  normal render walk, so steady-state cost is independent of tree size.


tomk-amd:
The dreaded AppWindow::WantsContinuousRender() - the side bar glow / pulse effect will not animate if you don't move the mouse.  Now this function will need to check the sidebar to see if it needs continuous rendering when a glow is active...  we will need to re-think how this continuous render logic should be implemented

src/view/src/rocprofvis_sidebar.cpp
```diff
@@ -365,8 +474,25 @@ void
 SideBar::RenderLeafNode(const LeafNode& leaf)
 {
     ImGui::PushID(static_cast<const void*>(&leaf));
+
+    const bool   is_reveal_target =
+        m_reveal_active && leaf.track_id == m_reveal_track_id;
```
drchen-amd, line 662: Queues may have multiple entries in the side bar, so there may be multiple nodes with the same track id. 

I'm seeing if a queue is included both in hw view as itself as well as in sw view under a stream, the instance that is scrolled to varies. If stream is collapsed, the hw view queue is scrolled to. If stream is not collapsed, the queue under stream is scrolled to.

Technically both are correct because they are both the same track. Just mentioning this inconsistency.

https://github.com/user-attachments/assets/7c0f3b3d-e6a6-43c7-a4eb-58ff6807cdb2

tomk-amd (approved):

PR #935, opened 2026-07-13
Refactor queries to make it optimal and correspond to new schema

amokiche-amd:
## Motivation
edit data flow to correspond new pc sampling schema

## Submission Checklist


src/model/src/database/rocprofvis_db_compute.h
```diff
@@ -1,290 +1,290 @@
-// Copyright Advanced Micro Devices, Inc.
```
drchen-amd, line 1: Are there any actual changes in here? If not, this file should be reverted.

src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -1,1644 +1,1687 @@
-// Copyright Advanced Micro Devices, Inc.
```
drchen-amd, line 1: This file should be reverted except for the actual changes.

src/model/inc/rocprofvis_interface_types.h
```diff
@@ -1,615 +1,615 @@
-// Copyright Advanced Micro Devices, Inc.
```
drchen-amd, line 1: This file should be reverted except for the actual changes.

drchen-amd (commented):
It is not clear to me what the intention is for how many pc sampling requests can be in flight at a time. Is it...

A) One pc sampling request per trace? 
B) One pc sampling request per kernel? (For trace for n kernels, there can be n requests in parllel)
C) One pc sampling request per kernel per source file? (For trace with n kernels each with m source files, there can be n x m requests in parellel).

- In code view, there is a single m_active_request_id that just gets cancelled whenever fetch happens, which suggest case A. 
- In DP, there is a guard that rejects the same kernel having multiple requests. Also, the requests output to persistent objects in each of the controller kernels. These two suggest case B.  
- The request ID it-self has kernel id and file id encoded inside, which suggest case C.

I suggest that all of these aspects are made to be in agreement.

src/model/inc/rocprofvis_interface_types.h
amokiche-amd: It something wrong with github compare, localy I see only real changes

Image

src/model/src/database/rocprofvis_db_compute.cpp
amokiche-amd: the same, locally all changes is ok

src/model/src/database/rocprofvis_db_compute.h
amokiche-amd: also github problem, localy all looks ok

amokiche-amd:
> I suggest that all of these aspects are made to be in agreement.

Agree, as I understood user can change anything in any case of time. The main idea is A - one request per trace etc it should close all requests except the latest. If user change the source file we close previous unfinished requests. And the same for the kernel changes. And now I realised I didn't process workload changes correctly. I will fix it, but main idea is above.

UPD: Fixed

src/model/inc/rocprofvis_interface_types.h
```diff
@@ -579,11 +581,12 @@ typedef enum rocprofvis_db_compute_use_case_enum_t
...
     kRPVComputeFetchKernelCodeObjects,
-    kRPVComputeFetchCodeObjectIsaLines,
-    kRPVComputeFetchIsaLineIsaLineDeps,
-    kRPVComputeFetchIsaLineSourceLineDeps,
-    kRPVComputeFetchIsaLineStallRecord,
-    kRPVComputeFetchStallRecordReasonCounts,
+    // Preserve public enum values after retiring the granular PC sampling queries.
+    kRPVComputeFetchKernelIsaToIsaDeps = 16,
+    kRPVComputeFetchKernelIsaLines = 20,
+    kRPVComputeFetchKernelIsaToSourceDeps,
```
tomk-amd: I don't think we need to preserve any values here - pc sampling api is still not public

tomk-amd (approved):
Clean up enum please, otherwise approved.

PR #938, opened 2026-07-17
Remove the COMPUTE_UI_SUPPORT feature flag

tomk-amd:
## Motivation

Compute support is a permament feature now.
Remove the COMPUTE_UI_SUPPORT feature flag.

### DIFF
--- a/src/controller/inc/rocprofvis_controller_enums.h
+++ b/src/controller/inc/rocprofvis_controller_enums.h
@@ -108,20 +108,18 @@ typedef enum rocprofvis_controller_object_type_t
     kRPVControllerObjectTypeEventArgument = 23,
     // Topology node object
     kRPVControllerObjectTypeTopologyNode = 24,
-#ifdef COMPUTE_UI_SUPPORT
     // Compute trace controller object
-    kRPVControllerObjectTypeControllerCompute = 100,
+    kRPVControllerObjectTypeControllerCompute = 25,
     // Workload object
-    kRPVControllerObjectTypeWorkload = 101,
+    kRPVControllerObjectTypeWorkload = 26,
     // Kernel object
-    kRPVControllerObjectTypeKernel = 102,
+    kRPVControllerObjectTypeKernel = 27,
     // Metrics container object
-    kRPVControllerObjectTypeMetricsContainer = 103,
+    kRPVControllerObjectTypeMetricsContainer = 28,
     // Roofline object
-    kRPVControllerObjectTypeRoofline = 104,
+    kRPVControllerObjectTypeRoofline = 29,
     //PcSampling
-    kRPVControllerObjectTypePCSampling = 105,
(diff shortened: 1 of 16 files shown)


src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -5067,6 +5052,8 @@ DataProvider::LoadRoofLineCeilingsCompute(WorkloadInfo&        workload,
     double   double_data = 0.0;
     uint64_t uint64_data = 0;
     uint64_t num_entries = 0;
+    (void) bandwidth_ridge;
```
drchen-amd: May as well remove the parameter from this function since it serves no purpose.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -5120,6 +5107,8 @@ DataProvider::LoadRoofLineCeilingsBandwidth(WorkloadInfo&        workload,
     double   double_data = 0.0;
     uint64_t uint64_data = 0;
     uint64_t num_entries = 0;
+    (void) compute_ridge;
```
drchen-amd: May as well remove the parameter from this function since it serves no purpose.

src/controller/inc/rocprofvis_controller_enums.h
```diff
@@ -108,20 +108,18 @@ typedef enum rocprofvis_controller_object_type_t
...
-    kRPVControllerObjectTypeKernel = 102,
+    kRPVControllerObjectTypeKernel = 27,
     // Metrics container object
-    kRPVControllerObjectTypeMetricsContainer = 103,
+    kRPVControllerObjectTypeMetricsContainer = 28,
     // Roofline object
-    kRPVControllerObjectTypeRoofline = 104,
+    kRPVControllerObjectTypeRoofline = 29,
     //PcSampling
-    kRPVControllerObjectTypePCSampling = 105,
```
drchen-amd, line 123: I liked having this "reserved" range for the compute objects. Prior to this they would become mixed in with the later systems trace additions. I just think it looks more neat if these compute/remote/profiler objects continue to be grouped in their own range.

drchen-amd (approved):

amokiche-amd (approved):

src/view/src/rocprofvis_data_provider.cpp
tomk-amd: Will do, I wasn't sure if they were perhaps kept for some future expansion / functionalitiy

src/controller/inc/rocprofvis_controller_enums.h
tomk-amd: Going to leave as is for now.. I would prefer to get rid of assignments all together .. but in keeping with the spirit of an ABI it is probably safer to leave them.

PR #939, opened 2026-07-17
Improve remote file browser

Sajeeth-Wimalasuriyan:
Current implementation is suboptimal this revamps the design using queues from Dipal

image

### DIFF
--- a/src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
+++ b/src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
@@ -1,3 +1,6 @@
+// Copyright Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
 #include "rocprofvis_controller_ssh_bridge.h"
 #include "rocprofvis_core_assert.h"
 
--- a/src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
+++ b/src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
@@ -9,14 +9,14 @@ namespace Controller
 SshBridge::SshBridge()
     : Handle(__kRPVControllerRemotePropertiesFirst,
              __kRPVControllerRemotePropertiesLast)
- 
+
 {
     m_ssh_status = kRPVControllerSshIdle;
 }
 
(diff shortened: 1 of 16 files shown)


src/view/src/remote/rocprofvis_ssh_test_dialog.cpp
```diff
@@ -74,6 +81,192 @@ namespace
...
+    }
+
+    // The trailing path component (basename) of a POSIX path.
+    std::string base_name(const std::string& path)
+    {
+        std::string::size_type pos = path.find_last_of('/');
+        return pos == std::string::npos ? path : path.substr(pos + 1);
+    }
+
+    // The extension (without dot) of a path's basename, "" if none.
```
tomk-amd: These should be moved to util files if they are general enough, and some are duplicates like lower_copy()

src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
```diff
@@ -1,390 +1,400 @@
...
-    // Clear any latched terminal status so a reused connection starts clean.
-    m_ssh_status = kRPVControllerSshIdle;
-}
-
-} // namespace Controller
-} // namespace RocProfVis
+#include "rocprofvis_controller_ssh_bridge.h"
+#include "rocprofvis_core_assert.h"
+
+namespace RocProfVis
```
tomk-amd, line 7: Since this whole file is being re-written, (white-space changes) please add the missing copyright lines to the top too..  and the header file:  rocprofvis_controller_ssh_bridge.h

Sajeeth-Wimalasuriyan: done

tomk-amd (approved):

tomk-amd:
Looks good but a lot of new magic numbers for sizing and spacing.

Also it would good to extract the actual file browser dialog into its own file.

src/controller/src/remote/rocprofvis_controller_ssh_bridge.cpp
tomk-amd, line 1: Only 14 lines are really changed in this file.. maybe it is the same line ending problem that Aleks saw - not sure why this is happening, as I've seen it in a few files too..

PR #941, opened 2026-07-17
Batch Track Settings

drchen-amd:
Adds the ability to modify track options for multiple tracks at once. The track options context menu gains two additional
  modes:

  1. **Apply to selected tracks** — Options common to all selected tracks are available. Any changes are propagated to all
  selected tracks.
  2. **Apply to all tracks of the same category** — Options belonging to the same type of track as the target track are
  available. Any changes are propagated to all tracks of the target's type.
     - Possible types are the available topology types: Queue / Instrumented Threads / Sampled Threads / Streams /
  Counters, with fallback to Event Track / Sample Track should there be no topology information.

  In both modes, the displayed state of options is the aggregate of all tracks that will be propagated to:

  - **Booleans**: Aggregated state is the AND of participating tracks.
  - **Enums**: Aggregated state is the same as participating tracks should they all share a common state, and empty if they
  do not.
  - **Others**: Aggregated state is the state of the last participating track.

  ## Changes

  - Added `TrackOptions` and moved ownership of user-configurable options out of `TrackItem`. Each `TrackItem` owns a
  `TrackOptions` through which it can access user options. Integration with `ProjectSettings` has also been moved into
  `TrackOptions`.
    - `TrackOptions` is a hierarchy consisting of `EventTrackOptions` / `CounterTrackOptions` / `QueueTrackOptions`. Their
  main purpose is to support option aggregation for the batch apply modes. Each type holds the unique options of its type plus render functions for the widgets that modify them.
      - `TrackOptions`: track visibility (`m_display`), track height (`m_height`)
      - `EventTrackOptions`: color mode (`m_color_mode`), compact mode (`m_compact`), expand mode (`m_expand`)
      - `CounterTrackOptions`: boxplot (`m_boxplot`), highlight (`m_highlight`), counter stats (`m_show_analysis`)
      - `QueueTrackOptions`: queue utilization (`m_show_queue_utilization`)
  - Added `TimelineTrackOptions` — manager for `TrackOptions`. `TrackItem`s delegate construction and initialization of
  `TrackOptions`, as well as handling of the context menu, to this class.
  - `Minimap` / `Histogram`
    - Decoupled updates from the `SideBar` and made them event driven.
      - `Minimap`: Updates on track reorder + track visibility toggle. Also fixes the `Minimap` recomputing data each frame
  when visible.
      - histogram: Updates on track visibility toggle.
    - Changed `UpdateHistogram(const std::vector<uint64_t>& interest_id, bool add)` to `UpdateHistogram(const
  std::vector<TrackItem*>& tracks)`.
      - Having the caller provide a list of track ids each time is cumbersome. Regardless of the list, all tracks are
  iterated to compute the histogram, so the list does not reduce the workload.
      - Its purpose is to update for track visibility, so just pass the track list instead, which carries visibility
  information.
  - Made `SideBar` eye updates event driven based on track visibility toggle.


src/view/src/rocprofvis_timeline_track_options.cpp
```diff
@@ -0,0 +1,1195 @@
...
+}
+
+EventTrackOptions::EventTrackOptions(const EventTrackOptions& other)
+: TrackOptions(other)
+, m_color_mode(other.m_color_mode)
+, m_compact(other.m_compact)
+, m_expand(other.m_compact)
+{
+    m_type_mask.set(TrackOptions::kEvent);
+}
```
tomk-amd, line 522: `m_expand(other.m_compact)` typo?

src/view/src/rocprofvis_timeline_track_options.cpp
```diff
@@ -0,0 +1,1195 @@
...
+            }
+            case Propagate::kSiblings:
+            {
+                if(m_options_aggregate_type)
+                {
+                    m_options_aggregate_type->Render();
+                }
+                break;
+            }
+        }
```
tomk-amd, line 998: switch without default will probably cause compiler warnings...

src/view/src/rocprofvis_timeline_track_options.cpp
```diff
@@ -0,0 +1,1195 @@
...
+            {
+                aggregate = std::make_unique<TrackOptions>(
+                    static_cast<TrackOptions&>(*components.front()));
+                break;
+            }
+            default:
+            {
+                ROCPROFVIS_ASSERT(true);
+                break;
+            }
```
tomk-amd, line 1049: if this is supposed to raise the assertion then it should be `ROCPROFVIS_ASSERT(false);`

src/view/src/rocprofvis_timeline_track_options.cpp
```diff
@@ -0,0 +1,1195 @@
...
+            {
+                options = std::make_unique<CounterTrackOptions>(
+                    track, *this, track.m_data_provider.GetTraceFilePath());
+                break;
+            }
+            default:
+            {
+                ROCPROFVIS_ASSERT(true);
+                break;
+            }
```
tomk-amd, line 872: if this is supposed to raise the assertion then it should be `ROCPROFVIS_ASSERT(false);`

src/view/src/rocprofvis_timeline_track_options.cpp
drchen-amd: Yes you're right. This was a last-minute addition. I'm still not sure whether to move this into trackoptions or keep it in trackitem. Unlike others in trackoptions, it is not in context menu nor is it part of project settings. But on the other hand it is a "track option".

I don't think this bug breaks anything because it would mean the aggregate expand state is wrong, but the aggregate expand state is not used because it is not part of context menu.

src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -135,29 +132,30 @@ FlameTrackItem::RenderMetaAreaExpand()
...
     if(visible_levels <= m_max_level + 1)
     {
         ImGui::SetCursorPos(button_pos);
         if(ImGui::ArrowButton("##expand", ImGuiDir_Down))
         {
             RecalculateTrackHeight();
-            m_is_expanded = true;
+            m_event_options->m_expand = true;
         }
         if(ImGui::IsItemHovered()) SetTooltipStyled("Expand track to see all events");
```
tomk-amd: should guard m_event_options? The `ROCPROFVIS_ASSERT(m_event_options);` in the constructor only triggers in debug builds.

ex:
```
if(m_event_options)
{
    m_event_options->m_expand = true;
}
```

There are other spots where this guard is already employed like line 198 below.

src/view/src/rocprofvis_track_item.cpp
```diff
@@ -157,6 +152,9 @@ TrackItem::TrackItem(DataProvider& dp, uint64_t id, bool display,
     m_selected_changed_token = EventManager::GetInstance()->Subscribe(
         static_cast<int>(RocEvents::kTimelineTrackSelectionChanged),
         selected_changed_handler);
+
+    m_options = m_timeline_track_options.InitTrack(*this);
+    ROCPROFVIS_ASSERT(m_options);
 }
```
tomk-amd, line 158: Same as other other assert.  This does not protect in release mode.  I think places where `m_options` is used should be guarded?

src/view/src/rocprofvis_flame_track_item.cpp
drchen-amd: I will add guards but I think the only case it will catch is track topology type being `Count`.

The case like line 198 was before I had fallback for unknown/uncategorized tracks. Previously these types would not construct `TrackOptions` so there was potential for null. Currently all tracks types (except for `Count` type) would construct TrackOptions.

tomk-amd (approved):

PR #944, opened 2026-07-20
Revamp the in-app profiler launcher UI/UX (rocprof-sys)

dhingora-amd:
## Motivation

The profiler launcher crammed every option, the command preview, and the live
output into one dense, flat dialog that was hard to scan and use. This PR
reworks it into a clean, card-based UI, reorganizes the rocprof-sys options so
the common path is simple, and splits configuration from execution. Scope is
the ROCm Systems Profiler (rocprof-sys) UI; the backend stays ready for
rocprof-compute and rocprofv3.

## Technical Details

- Two-step flow: a Configure view and a focused Run view (live output, elapsed
  timer, status, Cancel / Run Again / Back).
- Reorganized options: General keeps the essentials (preset, output format,
  trace window); everything else moves to a separate Advanced window
  (Sampling, ROCm, Perfetto, Process Sampling, Parallelism, Config & Logging).
- Modern, theme-aware widgets: card panels, toggle switches, a live config
  summary, and an Arguments & Environment panel using add-and-remove pills.
- Long lists render as responsive column grids; inputs are aligned and sized.
- Behavior: show the selected preset's description inline, remove the
  deprecated Timemory setting, and force PID-suffixed output off under MPI.
- Fixes/cleanup: restore standard window padding, clearer naming, and
  replace magic numbers / remove dead code.

### DIFF
--- a/src/view/src/profiler/rocprofvis_launch_shared_tabs.cpp
+++ b/src/view/src/profiler/rocprofvis_launch_shared_tabs.cpp
@@ -22,30 +22,22 @@ namespace
 
 static char s_save_preset_name[128] = {};
 
-// Design tokens for the custom launcher widgets below.
-constexpr float kCardRounding      = 10.0f;
-constexpr float kCardPadX          = 16.0f;
-constexpr float kCardPadY          = 13.0f;
-constexpr float kCardGap           = 7.0f;   // vertical gap between stacked cards
-constexpr float kAccentBarWidth    = 4.0f;   // card-header leading bar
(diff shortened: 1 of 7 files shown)


tomk-amd:
image

The new design is way too big.  Needlessly large padding.  Especially the bottom save / cancel area. Does not match style of other dialogs, like preferences.

image

Also large number of new magic numbers have been added, creating new padding and spacing "rules" not seen else where..
This should be made consistent with the rest of the application.

PR #945, opened 2026-07-20
Set up WIX v4 for Windows Installer

jbonnell-amd:
## Motivation

Adds support for generating a WIX installer using WIX v4 and a new `roc-optiq.wxs` script 

## Technical Details

- Updated `CMakeLists.txt` to add a new `PACKAGE_WIX` target for packaging the Windows installer
- Added `wix/roc-optiq.wxs` to handle the WIX installer generation for roc-optiq
- Updated `BUILDING.md` to include Windows Installer instructions

## Test Plan

- Ensure developers can generate the WIX installer locally
- Ensure no regressions with current builds/tests (WIX should only be required when building with `PACKAGE_WIX` target

## Test Result

- No build or test errors on existing CI (no WIX installed)
- When building locally with WIX installed, able to build installer (see BUILDING.md for reference)

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


wix/roc-optiq.wxs
```diff
@@ -0,0 +1,167 @@
...
+      BannerBmp  = absolute path to the 493x58 banner bitmap
+      DialogBmp  = absolute path to the 493x312 background bitmap
+      IconFile   = absolute path to ROCm_Optiq.ico
+-->
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs"
+     xmlns:ui="http://wixtoolset.org/schemas/v4/wxs/ui">
+
+  <Package Name="ROCm™ Optiq"
+           Manufacturer="Advanced Micro Devices, Inc."
+           Version="$(var.Version)"
```
amokiche-amd, line 27: in our cmake we use the 4 values for the vertions as:
set(PROJECT_VERSION_MAJOR 0)
set(PROJECT_VERSION_MINOR 5)
set(PROJECT_VERSION_PATCH 0)
set(PROJECT_VERSION_BUILD 1)

But wix use only 3 of them. It can be potential problem when user will try to upgrade one vertion to the second, because wix will ignore the changes.
For the solution we can concatenate the  PROJECT_VERSION_PATCH and PROJECT_VERSION_BUILD in one.
https://learn.microsoft.com/en-us/windows/win32/msi/productversion

wix/roc-optiq.wxs
```diff
@@ -0,0 +1,167 @@
...
+         Directory layout
+         Default: C:\Program Files\AMD\ROCm-Optiq\
+         ═══════════════════════════════════════════════════════════════════ -->
+    <StandardDirectory Id="ProgramFiles64Folder">
+      <Directory Id="AMDFolder" Name="AMD">
+        <Directory Id="INSTALLDIR" Name="ROCm-Optiq">
+
+            <Component Id="MainExecutable" Guid="*">
+              <File Id="RocOptiqExe"
+                    Name="roc-optiq.exe"
```
amokiche-amd, line 97: We have the DGLFW_LINK_STATIC flag in yaml scripts. I don't know is it real to assemble with DGLFW_LINK_STATIC=0 on windows, but we have this possibility and this is mean we should copy glfw.dll in the case if we build with DGLFW_LINK_STATIC=0.

wix/roc-optiq.wxs
```diff
@@ -0,0 +1,167 @@
...
+          <Shortcut Id="AppStartMenuShortcut"
+                    Name="ROCm™ Optiq"
+                    Description="ROCm™ Profiler Visualizer"
+                    Target="[INSTALLDIR]roc-optiq.exe"
+                    WorkingDirectory="INSTALLDIR"
+                    Icon="ApplicationIcon" />
+          <RemoveFolder Id="CleanupStartMenu"
+                        Directory="ApplicationProgramsFolder"
+                        On="uninstall" />
+          <RegistryValue Root="HKCU"
```
amokiche-amd: You use scope=perMachine, but all registry keys in the "current user" scope. Is it intentional? If not will be better use the HKLM branch of the registry.

wix/roc-optiq.wxs
```diff
@@ -0,0 +1,167 @@
...
+        <Shortcut Id="AppDesktopShortcut"
+                  Name="ROCm™ Optiq"
+                  Description="ROCm™ Profiler Visualizer"
+                  Target="[INSTALLDIR]roc-optiq.exe"
+                  WorkingDirectory="INSTALLDIR"
+                  Icon="ApplicationIcon" />
+        <RemoveFile Id="CleanupDesktopShortcut"
+                    Name="ROCm™ Optiq.lnk"
+                    On="uninstall" />
+        <RegistryValue Root="HKCU"
```
amokiche-amd: The same, looks like HKLM key much better

wix/roc-optiq.wxs
```diff
@@ -0,0 +1,167 @@
...
+        <Control Id="Cancel" Type="PushButton" X="304" Y="243" Width="56" Height="17" Cancel="yes" Text="!(loc.WixUICancel)">
+          <Publish Event="SpawnDialog" Value="CancelDlg" />
+        </Control>
+      </Dialog>
+
+      <!-- Wire InstallDirDlg → ShortcutOptionsDlg → VerifyReadyDlg -->
+      <Publish Dialog="InstallDirDlg" Control="Next" Event="NewDialog" Value="ShortcutOptionsDlg" Order="999" Condition="WIXUI_DONTVALIDATEPATH OR WIXUI_INSTALLDIR_VALID=&quot;1&quot;" />
+      <Publish Dialog="ShortcutOptionsDlg" Control="Back" Event="NewDialog" Value="InstallDirDlg" />
+      <Publish Dialog="ShortcutOptionsDlg" Control="Next" Event="NewDialog" Value="VerifyReadyDlg" />
+      <Publish Dialog="VerifyReadyDlg" Control="Back" Event="NewDialog" Value="ShortcutOptionsDlg" />
```
amokiche-amd: Minor, but user visible bug. 
wix has default 
Publish Dialog="VerifyReadyDlg"
         Control="Back"
         Event="NewDialog"
         Value="InstallDirDlg"
         Order="1"
         Condition="NOT Installed" 
And this event will have the bigger priority than current Publish Dialog="VerifyReadyDlg". That means when user will click back on the VerifyReadyDlg he will see the InstallDirDlg instead of the ShortcutOptionsDlg. (I checked it in assembled installer). As esiest fix you can add Order="999" for the VerifyReadyDlg

wix/roc-optiq.wxs
jbonnell-amd: Are you referring to the backend of WIX? At least from the .wxs script, the var.Version we provide from CMakeLists contains the full X.X.X.X version number (installer built locally is named roc-optiq-0.5.0.1-win64.msi).

jbonnell-amd: Ah nevermind, I looked into it some more and understand the issue now. I've updated it to merge the patch and build numbers together in the MSI package to better align with the MSI versioning specification. Thanks for pointing this out!

wix/roc-optiq.wxs
jbonnell-amd: Switched to HKLM. Thanks!

wix/roc-optiq.wxs
jbonnell-amd: Updated

wix/roc-optiq.wxs
jbonnell-amd: Good catch, I never noticed that when testing myself. Seems like adding Order=2 was sufficient to resolve this issue, so I just did that for now.

wix/roc-optiq.wxs
tomk-amd: Don't need to worry about this, we are always going to link glfw statically on production builds.

tomk-amd (approved):

PR #947, opened 2026-07-21
View: Refine inline device nodes in the sidebar topology tree

dhingora-amd:
## Motivation

In the timeline sidebar's topology tree, each stream lists the device it maps
to. That device node repeated the stream name (e.g. `stream 0 >>> GPU0`), which
was redundant and noisy, and the tracks nested under it competed visually with
the primary stream/device entries. This tidies up the inline device section so
the tree is easier to scan.

## Technical Details

- **Label dedup** (`rocprofvis_track_topology.cpp`): drop the `"<stream> >>> "`
  prefix from the inline device header so it reads just `GPU0` / `CPU0`.
- **Lead-in arrow** (`rocprofvis_sidebar.cpp`, `rocprofvis_tree_node.h`): a new
  `show_lead_arrow` flag on the inline device node draws a forward-arrow glyph
  (`ICON_ARROW_FORWARD`) between the collapse chevron and the label, so it reads
  `-> GPU0`. The label is padded with leading spaces to open the slot and the
  glyph is drawn at the label start; the chevron keeps its normal position and
  the tree-node ID stays bound to the label so collapse state is unaffected.
- **De-emphasize nested tracks** (`rocprofvis_sidebar.cpp`): dim the tracks
  under the device node (via `m_render_muted` + `ImGuiCol_TextDisabled`), scoped
  to the device node's children only — the device label itself stays normal.

### DIFF
--- a/src/view/src/rocprofvis_sidebar.cpp
+++ b/src/view/src/rocprofvis_sidebar.cpp
@@ -23,6 +23,9 @@ constexpr ImGuiTreeNodeFlags HEADER_FLAGS = ImGuiTreeNodeFlags_Framed |
 constexpr float TREE_LINE_W = 1.5f;
 constexpr float MENU_PAD_X  = 8.0f;
 constexpr float MENU_PAD_Y  = 6.0f;
+// ImGui offsets a framed tree node's label by FontSize + FramePadding.x * this
+// factor (see TreeNodeBehavior); used to place the inline device lead arrow.
+constexpr float FRAMED_LABEL_PAD_MULT = 3.0f;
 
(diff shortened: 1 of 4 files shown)


tomk-amd:
regarding: De-emphasize nested tracks (rocprofvis_sidebar.cpp): dim the tracks

After testing this I think we should drop this idea for now, at least in its current implementation.  When tracks are hidden they are also greyed out (de-enphasized) so we would need to choose another color or signifier.  Let's remove this for now.

tomk-amd (approved):

PR #949, opened 2026-07-22
Chrome trace format support

vstempen:
## Motivation

Make a working pilot of Perfetto trace formats support. It shows number of efforts on both sides (rorOptiq and potentially profiler hub) to support new format, rocOptic data model ability to ingest another format/schema adapter, controller and UI ability to handle missing data (mostly regarding topology implementation), etc.  

## Technical Details

In the initial commit the code will fail to build without Perfetto trace processor libraries prebuilt. 
Will try to add flexibility to skip the support in following commits.
Here are the libraries build instructions : [Building Perfetto Libraries](https://github.com/ROCm/roc-optiq/blob/a0a8509b5e1a9df0179aee69bf6b252b3c41a663/thirdparty/perfetto/trace_processor/build/README.md).
After libraries have been built, build the project and try to open Pytorch traces or any other Perfetto traces in .json, .proto or .pftrace formats. 

Also, in coming commits will reduce number of files. Those can be copied dynamically during build. Don't have to be submitted


drchen-amd:
Compared to branch point off main (e8beaeb826d8e3a0ac1b5d841ecae59f09cf8bd0), the histogram looks different:
Screenshot from 2026-07-27 18-05-53
image
image
image

(Left is this branch, right is e8beaeb826d8e3a0ac1b5d841ecae59f09cf8bd0)

drchen-amd:
Same with minimap, general shape is the same but the density shading is different:

Screenshot from 2026-07-27 18-23-21
Screenshot from 2026-07-27 18-23-49
Screenshot from 2026-07-27 18-23-57

vstempen:
Hate to say, but the new histogram is correct. We have a silent bug in main. The bug is in the histogram calculation we have (id + __op << 60) as event_id, but should be (id + (__op << 60)) as event_id because + has higher precedence in SQL than <<.

tomk-amd:
> Hate to say, but the new histogram is correct. We have a silent bug in main. The bug is in the histogram calculation we have (id + __op << 60) as event_id, but should be (id + (__op << 60)) as event_id because + has higher precedence in SQL than <<.

Yeah - playing with this branch the mini map behaviours feels more correct than on main

src/model/src/database/rocprofvis_db_trace_processor.h
```diff
@@ -0,0 +1,252 @@
...
+        // @param query - database query 
+        // @param file_path output path to write .CSV
+        // @param future - future object providing asynchronous execution mechanism 
+        // @return status of operation
+        rocprofvis_dm_result_t ExportTableCSV(
+            rocprofvis_dm_charptr_t query,
+            rocprofvis_dm_charptr_t file_path,
+            Future* future) override {
+            return kRocProfVisDmResultSuccess;
+        };
```
drchen-amd: Does this mean export is not supported for this db type? If so this can rely on default implementation in base `Database` and return not supported.

src/model/src/database/rocprofvis_db_trace_processor.h
```diff
@@ -0,0 +1,252 @@
...
+        rocprofvis_dm_result_t  ReadExtEventInfo(
+            rocprofvis_dm_event_id_t event_id,
+            Future* object) override ;
+
+        rocprofvis_dm_result_t SaveTrimmedData(rocprofvis_dm_timestamp_t start,
+            rocprofvis_dm_timestamp_t end,
+            rocprofvis_dm_charptr_t new_db_path,
+            Future* future) override {
+            return kRocProfVisDmResultSuccess;
+        };
```
drchen-amd, line 89: Does this mean this db type does not support trimming? Should return not supported if so.

src/model/src/database/rocprofvis_db_cache.h
```diff
@@ -61,18 +61,18 @@ class TableCache
...
-        void AddTableCell(const char* table_name, uint64_t row_id, const char* column_name, rocprofvis_db_data_type_t column_type, const char* cell_value);
-        void AddTableRow(const char* table_name, uint64_t row_id);
-        void AddTableColumn(const char* table_name, const char* column_name, rocprofvis_db_data_type_t column_type);
-        const char* GetTableCell(const char* table_name, uint64_t row_id, const char* column_name);
-        const char* GetTableCellByIndex(const char* table_name, uint32_t row_index, const char* column_name);
-        void* GetTableHandle(const char* table_name);
+        void AddTableCell(std::string table_name, uint64_t row_id, uint32_t column_index, std::string cell_value);
+        void AddTableCell(std::string table_name, uint64_t row_id, std::string column_name, rocprofvis_db_data_type_t column_type, std::string cell_value);
+        void AddTableRow(std::string table_name, uint64_t row_id);
+        void AddTableColumn(std::string table_name, std::string column_name, rocprofvis_db_data_type_t column_type);
```
drchen-amd: Is there a benefit of switching these `Add` functions to std::string? Ultimately they are stored as std::string in `row` vector or `tables` map so the end result does not change, but I think previous const char* saves us an intermediate copy.

drchen-amd (commented):
Existing functionality seems otherwise consistent with main on ubuntu for rpd, db, and compute. Only tested building on windows.

src/model/src/database/rocprofvis_db_cache.h
tomk-amd: can we change the interface to pass by reference, ex: const std::string& table_name, ... ?

src/model/src/database/rocprofvis_db_trace_processor.h
vstempen: Oh, that I just missed. Will add it.

vstempen: Done

src/model/src/database/rocprofvis_db_cache.h
vstempen: There was a reason, but it could have been handled different way. This is valid concern. Back to const char*.

src/controller/src/system/rocprofvis_controller_trace_system.cpp
```diff
@@ -577,7 +577,14 @@ rocprofvis_result_t SystemTrace::Load(RocProfVis::Controller::Future& future)
...
                 m_trace_file.find(".db", m_trace_file.size() - 3) != std::string::npos ||
-                m_trace_file.find(".yaml", m_trace_file.size() - 5) != std::string::npos)
+                m_trace_file.find(".yaml", m_trace_file.size() - 5) != std::string::npos 
+#ifdef ROCPROFVIS_PERFETTO_ENABLED
+                ||
+                m_trace_file.find(".json", m_trace_file.size() - 5) != std::string::npos ||
+                m_trace_file.find(".proto", m_trace_file.size() - 6) != std::string::npos || 
+                m_trace_file.find(".pftrace", m_trace_file.size() - 8) != std::string::npos
+#endif
+                )
```
tomk-amd: Suggestion: https://github.com/ROCm/roc-optiq/pull/969

Simplify this without using magic numbers

tomk-amd (approved):

PR #950, opened 2026-07-22
Roofline: single-click filtering for kernels, memory levels, and bandwidth peaks

dhingora-amd:
## Motivation

The roofline shows every kernel's dot for every memory level plus all
bandwidth ceilings at once, which is noisy and takes many clicks to pare
down. This PR adds single-click ways to focus the view; they compose and
don't change the default.

## Technical Details

All in `src/view/src/compute/rocprofvis_compute_roofline.{h,cpp}`:

1. **Kernel-intensity filter** – Options dropdown (All / HBM / L2 / L1 / LDS)
   limits kernel dots to one memory level.
2. **Kernel isolation** – click a kernel's dot or legend row to show only that
   kernel; click again to restore all.
3. **Bandwidth-peak isolation** – click a bandwidth row or its plot line to show
   only that peak; the legend keeps all rows and highlights the active one.

Filters apply to the plot, legend, and hover; plot clicks use a drag check so
panning isn't treated as a click; state resets on workload change.

### DIFF
--- a/src/view/src/compute/rocprofvis_compute_roofline.cpp
+++ b/src/view/src/compute/rocprofvis_compute_roofline.cpp
@@ -70,6 +70,7 @@ Roofline::Roofline(DataProvider& data_provider, KernelMode kernel_mode)
 , m_menus_placement(InsideTopRight)
 , m_scale_intensity(true)
 , m_line_thickness(LINE_THICKNESS_DEFAULT)
+, m_memory_peak_filter(std::nullopt)
 , m_menus_rendered_height(0.0f)
 , m_hovered_item_distance(FLT_MAX)
 , m_workload_changed(false)
@@ -436,7 +437,8 @@ Roofline::Render()
                     {
                         display =
                             m_items[i].info.intensity &&
-                            (m_kernel ? m_items[i].parent_info.kernel == m_kernel : true);
+                            (m_kernel ? m_items[i].parent_info.kernel == m_kernel : true) &&
+                            IntensityMatchesMemoryFilter(m_items[i]);
                     }
                 }
                 display &= m_items[i].visible;
@@ -787,7 +789,7 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
         ImGui::EndGroup();
         float header_height = ImGui::GetItemRectSize().y + 2 * style.WindowPadding.y;
         float footer_height =
-            (m_kernel_mode == AllKernels ? 6 : 5) * ImGui::GetFrameHeightWithSpacing() +
+            (m_kernel_mode == AllKernels ? 8 : 7) * ImGui::GetFrameHeightWithSpacing() +
             2 * style.WindowPadding.y;
         ImGui::SetNextWindowSizeConstraints(
             ImVec2(menus_content_width, 0),
@@ -820,7 +822,8 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
                 {
                     display =
                         m_items[i].info.intensity &&
-                        (m_kernel ? m_items[i].parent_info.kernel == m_kernel : true);
+                        (m_kernel ? m_items[i].parent_info.kernel == m_kernel : true) &&
+                        IntensityMatchesMemoryFilter(m_items[i]);
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -1059,5 +1135,59 @@ Roofline::ApplyPreset(PresetModel::Type type)
...
+Roofline::KernelPassesIsolation(const ItemModel& item) const
+{
+    bool passes = true;
+    if(m_kernel_mode == AllKernels && m_isolated_kernel &&
+       item.type == ItemModel::Type::Intensity)
+    {
+        passes = item.parent_info.kernel == m_isolated_kernel;
+    }
+    return passes;
+}
```
drchen-amd: When I click a kernel, it gets removed in the chart and legend, but it also gets removed in the menus. The menus should not be affected, it should always have the entire data set.

I suggest taking a look at PresetModel and ApplyPresets. PresetModel is just indexes of plot items that is gathered when data updates. Each plot item has `visible` that controls whether they show up in plot and legend. ApplyPresets takes the indexes from the presets and writes `visible` accordingly, and I think this is the only thing all these filters/isolators need to do as well.

src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -1059,5 +1135,59 @@ Roofline::ApplyPreset(PresetModel::Type type)
...
+{
+    if(m_isolated_bandwidth && m_isolated_bandwidth.value() == bandwidth)
+    {
+        m_isolated_bandwidth = std::nullopt;
+    }
+    else
+    {
+        m_isolated_bandwidth = bandwidth;
+    }
+}
```
drchen-amd: When you change ceilings, you have to set `m_options_changed`. Otherwise you will get discontinuities.

drchen-amd:
Controls should be moved under presets header. Instead of the 6 buttons we have now, there should be a dropdown for each of the filters you added

1. Kernel-intensity filter, options: All / HBM / L2 / L1 / LDS.
2. Kernel isolation, options: names of all kernels
3. Bandwidth-peak isolation: options: All / HBM / L2 / L1 / LDS

The click shortcut for 2 & 3 can stay to compliment the dropdowns since they are not easily discoverable by themselves.

src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -912,6 +961,32 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
...
+                       plot_size.x * 0.5f, Alignment_Left, true);
+            ImGui::SetNextItemWidth(-1.0f);
+            int intensity_filter_idx =
+                m_memory_peak_filter
+                    ? static_cast<int>(m_memory_peak_filter.value()) + 1
+                    : 0;
+            PushComboStyles();
+            // Order must match the intensity enum: All, HBM, L2, L1, LDS.
+            if(ImGui::Combo("##memory_peak", &intensity_filter_idx,
+                            "All\0HBM\0L2\0L1\0LDS\0\0"))
```
drchen-amd: These may be empty. Option should only be available for non-empty intensities.

drchen-amd:
Theres also this weird behavior where the legend highlight sticks if clicked, but not the first 2 items...

[Screencast from 07-22-2026 07:56:29 PM.webm](https://github.com/user-attachments/assets/74217717-ee22-4338-97b7-65a1b21ea913)

Previously clicking the legend did not persist the highlighting.

src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -771,16 +837,129 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
...
+                if(ImGui::BeginCombo("##kernel",
+                                     m_isolated_kernel ? m_isolated_kernel->name.c_str()
+                                                       : FILTER_OPTION_ALL))
+                {
+                    if(ImGui::Selectable(FILTER_OPTION_ALL, !m_isolated_kernel))
+                    {
+                        m_isolated_kernel = nullptr;
+                        RecomputeVisibility();
+                    }
+                    for(const KernelInfo* kernel : m_workload->ordered_kernels)
```
drchen-amd: Some kernels do not have roofline data (`kernel.roofline.intensities.empty()`) and should not be in this list.

src/view/src/compute/rocprofvis_compute_roofline.cpp
```diff
@@ -792,7 +971,7 @@ Roofline::RenderMenus(ImVec2 region, ImVec2 plot_pos, ImVec2 plot_size,
         ImGui::SetNextWindowSizeConstraints(
             ImVec2(menus_content_width, 0),
             ImVec2(menus_content_width,
-                   max_menus_height - header_height - footer_height));
+                   std::max(0.0f, max_menus_height - header_height - footer_height)));
```
drchen-amd: Is this needed? It breaks the custom region when the height is small. 

Screenshot from 2026-07-23 19-15-32

drchen-amd:
I think the order and names of the dropdowns should be changed to:
1. "Preset" -> "Compute peak"
2. "Bandwidth peak"
3. "Kernel"
4. "Kernel intensity" ->"Kernel bandwidth"

Thoughts @tomk-amd?

tomk-amd:
> I think the order and names of the dropdowns should be changed to:
> 
> 1. "Preset" -> "Compute peak"
> 2. "Bandwidth peak"
> 3. "Kernel"
> 4. "Kernel intensity" ->"Kernel bandwidth"
> 
> Thoughts @tomk-amd?

I think this entire interface needs to be rethought ...

image

First laying out these quick use pull downs vertically make this panel very tall - requires scrolling - also in smaller views (as mentioned already) the custom portion get clipped and is unusable.

Second these are supposed to be quick options, but being buried behind the settings (gear) button does not make them quick to access.  These options should always be accessible..  One option is to lay them out horizontally at the top like a tool bar?

I think the only real options, things that should be in this settings panel, are: line thickness, legend position, and the scale marker checkbox.

drchen-amd:
Yeah I think something like this would work:
image

Although horizontal layout will mean it will handle small window sizes less gracefully than vertical laying out. I imagine it will be like the compute comparison tool bar where, less than some minimum width, we give up and spill off screen. At least with vertical layout, scrolling can accommodate any window size.

With regards to custom portion getting clipped, if that line I called out is reverted, custom should never clip. It is not obvious to me the purpose of that change.

drchen-amd:
When you make changes via "custom", the dropdowns will no longer reflect the displayed data. I'm not sure if this okay, or potentially misleading.

image

If this is undesirable, we could do something simple like whenever a change is made in "custom", all the dropdowns are set to a blank option "-".

drchen-amd (approved):
(I pushed a change to fix the legend not fully utilizing available space. Its not related to your changes, seems like existing issue.)

PR #791, opened 2026-04-21
Enable ImGui docking + multi-viewport

Sajeeth-Wimalasuriyan:
image

PR #952, opened 2026-07-23
Dipal and Sajeeth Combined Remote Profile Redesign 

Sajeeth-Wimalasuriyan:
image
image
image


amokiche-amd:
Great improvements, looks very great. A small suggestion regarding the profiler window. I find it odd that it appears as a window inside the main window and can be moved, but only within the main window. I think it would be better to make it fixed, so that it occupies the entire space of the main window and cannot be moved. But if it needs to be movable, then it probably makes sense to make it a fully independent window that can be moved outside of the main window.

tomk-amd:
> then it probably makes sense to make it a fully independent window that can be moved outside of the main window.

This is the final goal, if we ever have the technology :) https://github.com/ROCm/roc-optiq/pull/791

tomk-amd (approved):

PR #956, opened 2026-07-24
View: add RenderScheduler to drive lazy-render wake-ups

dhingora-amd:
## Motivation

The lazy render loop sleeps when idle, and every feature that animates had to
add its own branch to `WantsContinuousRender()` to stay awake. That doesn't
scale and is easy to miss — the pulsing event highlight was never wired in, so
it stuttered when the mouse was idle.

## Technical Details

Add `RenderScheduler`: subsystems push a frame request from their own per-frame
code instead of `WantsContinuousRender()` polling each one. Requests are
level-triggered and self-expiring (nothing to release).

- Fix idle pulse stutter (`TimelineSelection` requests frames while a transient
  highlight is active).
- Migrate notifications, log viewer, app monitor, and timeline
  (loading debounce / sticky drag / reorder) to push; remove
  `RootView`/`TraceView`/`TimelineView::WantsContinuousRender()`.
- `WantsContinuousRender()` now checks `RenderScheduler` plus the foundational
  conditions and the cross-tab data-provider load poll.

### DIFF
--- a/src/view/src/rocprofvis_appmonitor.cpp
+++ b/src/view/src/rocprofvis_appmonitor.cpp
@@ -4,6 +4,7 @@
 #include "rocprofvis_appmonitor.h"
 #include "rocprofvis_event_manager.h"
 #include "rocprofvis_controller.h"
+#include "rocprofvis_render_scheduler.h"
 
 #include <spdlog/spdlog.h>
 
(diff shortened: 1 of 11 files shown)


Sajeeth-Wimalasuriyan:
Looks good will wait for tom to give the final OK

tomk-amd (approved):

PR #957, opened 2026-07-24
Remote Profiling Mini Fixes

Sajeeth-Wimalasuriyan:
## Summary
Addresses review feedback on the remote panel revamp. No behavior changes
beyond the noted robustness fixes; all theming/dedup work is cosmetic-neutral.
## Changes
- **Themed warning color:** Added `Colors::kTextWarning` (dark + light theme
  entries) and routed the launcher readiness label plus backend warning
  messages through the theme instead of hardcoded amber/red/blue literals.
- **De-duplicated download popup:** Extracted the copy-pasted remote-download
  progress modal into a shared `RenderRemoteDownloadPopup` helper in
  `gui_helpers`, now used by both `SshTestDialog` and the profiler launcher.
- **Brace style:** Added braces to brace-less control flow in the remote file
  browser, shared launch tabs, launcher dialog, and the SSH client realpath
  block (per `CODING.md`).
- **More themed literals:** `StatusPill` text and the toggle-switch knob now use
  `Colors::kTextOnAccent`; `RenderOutputConsole` uses the shared
  `ChildRounding` instead of a hardcoded `8.0f`.
- **SFTP null-handle guard:** Bail out of the directory listing when `opendir`
  returns a null handle, guarding `realpath`/`readdir`/`closedir` against a null
  dereference (also fixes a latent case where a network-failed listing returned
  `Success`).
- **Safer profile rehydrate:** `LoadFromSettings` now validates the profile's
  SSH connection ref against the connection store (matching the toolbar load
  path), so a deleted connection is left unbound instead of silently rebound.

### DIFF
--- a/src/controller/src/remote/rocprofvis_controller_ssh_client.cpp
+++ b/src/controller/src/remote/rocprofvis_controller_ssh_client.cpp
@@ -1184,36 +1184,26 @@ namespace Controller
         return Result::Success;
     }
 
-    // libssh2's SFTP teardown entry points can report EAGAIN on a non-blocking
-    // session, so releasing a handle needs the same bounded socket wait as the
-    // rest of the transport. These guards own that retry loop, so a listing
-    // releases its handles exactly once on every exit path - success, error or
-    // cancel - without repeating the teardown at each return. WaitSocket is
-    // injected as a function pointer to keep the guards local to this file.
-    using wait_socket_fn_t = bool (*)(SshConnection*);
-
+    // SFTP teardown can report EAGAIN on a non-blocking session, so closing a
+    // handle needs the same bounded socket wait as the rest of the transport.
     class SftpSessionGuard
     {
     public:
-        SftpSessionGuard(SshConnection* connection, LIBSSH2_SFTP* sftp,
-                         wait_socket_fn_t wait_socket)
+        SftpSessionGuard(SshConnection* connection, LIBSSH2_SFTP* sftp)
             : m_connection(connection)
             , m_sftp(sftp)
-            , m_wait_socket(wait_socket)
         {
         }
 
         ~SftpSessionGuard()
         {
-            if (m_sftp != nullptr)
+            while (m_sftp != nullptr &&
+                   libssh2_sftp_shutdown(m_sftp) == LIBSSH2_ERROR_EAGAIN)
             {
-                while (libssh2_sftp_shutdown(m_sftp) == LIBSSH2_ERROR_EAGAIN)
+                if (!SshClient::WaitSocket(m_connection))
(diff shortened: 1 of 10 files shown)


src/controller/src/remote/rocprofvis_controller_ssh_client.cpp
amokiche-amd: If we brake the loop here the sftp will be null, but we continue to execute this function and call libssh2_sftp_opendir(sftp, path.c_str())) with null and libssh2_sftp_shutdown(sftp) with null. It looks like we should make a return in this case, or check this variable before use

amokiche-amd: The same as above

src/view/src/profiler/rocprofvis_profiler_launcher_dialog.cpp
```diff
@@ -1596,9 +1528,15 @@ void ProfilerLauncherDialog::LoadFromSettings()
                 }
             }
 #ifdef ROCPROFVIS_ENABLE_REMOTE
-            if (!m_config.ssh_connection_ref.empty())
+            // Only rebind the remote section when the profile references a
+            // connection that still exists; a stale ref (deleted connection) is
+            // left unbound rather than silently mapped to another connection.
+            // Mirrors the combo-select load path in RenderToolbar().
+            if (!m_config.ssh_connection_ref.empty() &&
+                m_connection_store.Get(m_config.ssh_connection_ref) != nullptr)
```
amokiche-amd, line 1537: As I understood right here m_connection_store.Load() not called and it will be always empty and it if will be always falls and body won't be executed

src/controller/src/remote/rocprofvis_controller_ssh_client.cpp
amokiche-amd: Also, we report error, but continue execution.

amokiche-amd: Also, we report error, but can exit the function with Success status. Even if it intentional looks strange.

amokiche-amd: Each path with success or error in this function should call libssh2_sftp_shutdown(sftp) (the same for the LIBSSH2_SFTP_HANDLE* dir should be called close). In this function libssh2_sftp_shutdown called 3 times and it easy in next changes forget to close it. I think it will be better wrap it in RAII object (also for dir variable). Yoy get rid of manual call of libssh2_sftp_shutdown and will be defenetly sure it will be called whe execution leve the function.

tomk-amd (approved):

PR #958, opened 2026-07-24
Table API Refactor

drchen-amd:
This change split the table api into event search and "regular" paths while maintaining code reuse for common functionality. This will allow these use cases to be independently expanded without changes spilling over to the other. There are no functional changes.

### DM
- Moved search specific parameters of `rocprofvis_db_build_table_query` into new API `rocprofvis_db_build_event_search_query`
- `BuildTableQuery`:
  - Previous implementation had 2 parts; building the SQL; building the CMD.
    - "Regular" SQL was duplicate of `BuildSliceQueryMap`, but with the `kRPVQueryTable` query type. Replaced with call to BuildSliceQueryMap with added query type parameter. 
    - Search specific SQL moved into new function `BuildEventSearchQuery`
    - CMD moved into new function `BuildCompoundQuery`
- Flow: 
  - `rocprofvis_db_build_table_query` -> `BuildTableQuery` -> `BuildSliceQueryMap` + `BuildCompoundQuery`
  - `rocprofvis_db_build_event_search_query` -> `BuildEventSearchQuery` -> `BuildSliceQueryMap` + inject search filter + `BuildCompoundQuery` 
- Removed unused analysis `TableProcessor` (leftover from queue utilization V2).

### Controller
- `Table`
  - Added `SystemTable:EventSearchTable`
  - Redistributed existing members among hierarchy where applicable.
  - Virtualized all input arguments and areas that deal with arguments. This is so base classes can operate on generic argument objects while specific handling is routed to appropriate derives.
    - `TableArguments`, `SystemTableArguments:TableArguments`,  `EventSearchTableArguments:SystemTableArguments`

### View
- Added `TrackTableRequestParams:TableRequestParams`, `EventSearchRequestParams:TableRequestParams`
- Redistributed logic of `FetchTable` into newly added `FetchTrackTable`, `FetchEventSearch`.
- Reverted misleading search bar hint text.

### DIFF
--- a/src/controller/src/rocprofvis_controller_analysis.cpp
+++ b/src/controller/src/rocprofvis_controller_analysis.cpp
@@ -450,8 +450,13 @@ Analysis::EventsTable::EventsTable(uint64_t id, rocprofvis_dm_event_operation_t 
 , m_op(op)
 {}
 
-rocprofvis_result_t Analysis::EventsTable::UnpackArguments(Arguments& args, QueryArguments& out) const
+rocprofvis_result_t Analysis::EventsTable::UnpackArguments(Arguments& args, TableArguments*& out) const
 {
+    if(!out)
+    {
+        out = new SystemTableArguments();
+    }
+    SystemTableArguments* sys_out = (SystemTableArguments*)out;
     rocprofvis_result_t result = kRocProfVisResultInvalidArgument;
     std::vector<uint32_t> tracks;
     uint64_t num_tracks = 0;
@@ -494,13 +499,21 @@ rocprofvis_result_t Analysis::EventsTable::UnpackArguments(Arguments& args, Quer
             }
         }
     }
-    const char* group = (m_op == kRocProfVisDmOperationLaunchSample) ? "name, COUNT(*) AS Invocations, SUM(duration) AS DurationTotal" :
-        "name, COUNT(*) AS Invocations, SUM(duration) AS DurationTotal, AVG(duration) AS DurationAvg, MIN(duration) AS DurationMin, MAX(duration) AS DurationMax";
     ROCPROFVIS_ASSERT(result == kRocProfVisResultSuccess);
     result = args.GetUInt64(kRPVControllerTableArgsSortColumn, 0, &sort_column_index);
     ROCPROFVIS_ASSERT(result == kRocProfVisResultSuccess);
     result = args.GetUInt64(kRPVControllerTableArgsSortOrder, 0, &sort_order);
-    out = {"", "__op = " + std::to_string(m_op), group, "name", sort_column_index, (rocprofvis_controller_sort_order_t)sort_order, tracks, {}, use_case, start_ts, end_ts};
+    sys_out->m_sort_column = sort_column_index;
+    sys_out->m_sort_order = (rocprofvis_controller_sort_order_t)sort_order;
+    sys_out->m_where = "";
+    sys_out->m_filter = "__op = " + std::to_string(m_op);
+    sys_out->m_group = (m_op == kRocProfVisDmOperationLaunchSample) ? "name, COUNT(*) AS Invocations, SUM(duration) AS DurationTotal" :
+        "name, COUNT(*) AS Invocations, SUM(duration) AS DurationTotal, AVG(duration) AS DurationAvg, MIN(duration) AS DurationMin, MAX(duration) AS DurationMax";
+    sys_out->m_group_cols = "name";
+    sys_out->m_tracks = std::move(tracks);
+    sys_out->m_use_case = use_case;
+    sys_out->m_start_ts = start_ts;
+    sys_out->m_end_ts = end_ts;
     return result;
 }
 
--- a/src/controller/src/rocprofvis_controller_table.h
+++ b/src/controller/src/rocprofvis_controller_table.h
@@ -30,6 +30,8 @@ public:
 
     virtual rocprofvis_controller_object_type_t GetType(void);
 
+    virtual void Reset();
+
     virtual rocprofvis_result_t SetupAndFetch(Trace& controller, Arguments& args, Array& array, Future* future);
     virtual rocprofvis_result_t ExportCSV(rocprofvis_dm_trace_t dm_handle, Arguments& args, Future* future, const char* path) const = 0;
 
@@ -39,16 +41,21 @@ protected:
         std::string m_name;
         rocprofvis_controller_primitive_type_t m_type;
     };
+    struct TableArguments
+    {
+        virtual ~TableArguments() = default;
+        uint64_t m_sort_column;
+        rocprofvis_controller_sort_order_t m_sort_order;
+    };
 
     virtual rocprofvis_result_t Setup(rocprofvis_dm_trace_t dm_handle, Arguments& args, Future* future) = 0;
     virtual rocprofvis_result_t Fetch(rocprofvis_dm_trace_t dm_handle, uint64_t index, uint64_t count, Array& array, Future* future) = 0;
+    virtual rocprofvis_result_t UnpackArguments(Arguments& args, TableArguments*& out) const;
+    virtual void                GetCurrentArguments(TableArguments*& out) const;
+    virtual void                SetCurrentArguments(TableArguments& in);
 
     std::vector<ColumnDefintion> m_columns;
     std::map<uint64_t, std::vector<Data>> m_rows;
-    std::string m_where;
-    std::string m_filter;
-    std::string m_group;
-    std::string m_group_cols;
     uint64_t m_num_items;
     uint64_t m_id;
     uint64_t m_sort_column;
--- a/src/controller/src/system/rocprofvis_controller_table_system.cpp
+++ b/src/controller/src/system/rocprofvis_controller_table_system.cpp
@@ -31,21 +31,15 @@ SystemTable::~SystemTable()
 
 void SystemTable::Reset()
 {
+    Table::Reset();
     m_start_ts = 0;
     m_end_ts = 0;
-    m_num_items = 0;
-    m_sort_column = 0;
-    m_sort_order = kRPVControllerSortOrderAscending;
     m_use_case = kRPVDMTableUseCaseEventTrackTable;
-    m_columns.clear();
-    m_rows.clear();
     m_tracks.clear();
     m_where.clear();
     m_filter.clear();
     m_group.clear();
     m_group_cols.clear();
-    m_string_table_filters.clear();
-    m_string_table_filters_ptr.clear();
 }
 
 rocprofvis_result_t SystemTable::SetupAndFetch(Trace& controller, Arguments& args, Array& array, Future* future)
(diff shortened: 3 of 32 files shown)


src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -501,62 +453,55 @@ rocprofvis_result_t SystemTable::GetString(rocprofvis_property_t property, uint6
...
-
-    result = UnpackUseCase(args, use_case);
-    if (result == kRocProfVisResultSuccess)
+    if(!out)
     {
-        result = args.GetUInt64(kRPVControllerTableArgsType, 0, &table_type);
+        out = new SystemTableArguments();
     }
-    if (result == kRocProfVisResultSuccess)
+    SystemTableArguments* sys_out = (SystemTableArguments*)out;
```
amokiche-amd, line 462: C-style cast is unsafe, use dynamic_cast here will be better

src/model/src/database/rocprofvis_db.cpp
```diff
@@ -159,6 +159,30 @@ rocprofvis_dm_result_t   Database::ReadEventPropertyAsync(
     return kRocProfVisDmResultSuccess;
 }
 
+rocprofvis_dm_result_t Database::BuildEventSearchQuery(
```
amokiche-amd, line 162: What the sense make this stub for base class? Class database already has the pure virtual functions and cant have the objects. It will be better to make this function also pure virtual.

src/controller/src/system/rocprofvis_controller_table_system_search.cpp
```diff
@@ -0,0 +1,138 @@
...
+    {
+        uint64_t num_string_table_filters = 0;
+        std::vector<std::string> string_table_filters;
+        std::vector<const char*> string_table_filters_ptr;
+        uint64_t table_type = static_cast<uint64_t>(kRPVControllerTableTypeSearchResults);
+
+        if (result == kRocProfVisResultSuccess)
+        {
+            result = args.GetUInt64(kRPVControllerTableArgsType, 0, &table_type);
+        }
```
amokiche-amd: What the point of this check? We already inside the if above and did nothing with result variable

src/controller/src/rocprofvis_controller_table.h
```diff
@@ -39,16 +41,21 @@ class Table : public Handle
         std::string m_name;
         rocprofvis_controller_primitive_type_t m_type;
     };
+    struct TableArguments
+    {
+        virtual ~TableArguments() = default;
```
amokiche-amd, line 46: There is no point to make this destructor virtual

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -67,104 +61,108 @@ rocprofvis_result_t SystemTable::Fetch(rocprofvis_dm_trace_t dm_handle, uint64_t
...
-        char const* sort_column = m_columns.size() > m_sort_column ? m_columns[m_sort_column].m_name.c_str() : nullptr;
-
-        char* fetch_query = nullptr;
-        rocprofvis_dm_result_t dm_result = rocprofvis_db_build_table_query(
-            db, m_use_case, static_cast<rocprofvis_dm_timestamp_t>(m_start_ts), static_cast<rocprofvis_dm_timestamp_t>(m_end_ts),
-            static_cast<rocprofvis_db_num_of_tracks_t>(m_tracks.size()), m_tracks.data(), m_where.c_str(),  m_filter.c_str(), m_group.c_str(), m_group_cols.c_str(), sort_column,
-            (rocprofvis_dm_sort_order_t)m_sort_order, static_cast<rocprofvis_dm_num_string_table_filters_t>(m_string_table_filters_ptr.size()), m_string_table_filters_ptr.data(), 
-            count, index, false, &fetch_query);
-        rocprofvis_dm_table_id_t table_id = 0;
+        TableArguments* fetch_args = nullptr;
```
amokiche-amd, line 64: Smart pointers like std::unique_ptr looks better and safer here than old C-style raw pointers. You lose nothing, but win with safety in potential changes.

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -607,95 +551,134 @@ SystemTable::UnpackArguments(Arguments& args, QueryArguments& out) const
...
+            sys_out->m_start_ts = start_ts;
+            sys_out->m_end_ts = end_ts;
+        }   
     }
-    if(num_op_types > 0 && result == kRocProfVisResultSuccess)
+	return result;
+}
+
+void
+SystemTable::GetCurrentArguments(TableArguments*& out) const
```
amokiche-amd, line 629: By the modern c++ guidlines:
F.20: For “out” output values, prefer return values to output parameters
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?utm_source=chatgpt.com#rf-out

It easy to use here

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -607,95 +551,134 @@ SystemTable::UnpackArguments(Arguments& args, QueryArguments& out) const
...
+{
+    if(!out)
     {
-        result = args.GetUInt64(kRPVControllerTableArgsNumStringTableFilters, 0, &num_string_table_filters);
-        if(result == kRocProfVisResultSuccess)
+        out = new SystemTableArguments();
```
amokiche-amd, line 633: make it unique_ptr and use make_unique insted of call new manualy.

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -607,95 +551,134 @@ SystemTable::UnpackArguments(Arguments& args, QueryArguments& out) const
...
+    }
+    SystemTableArguments* sys_out = (SystemTableArguments*)out;
+    Table::GetCurrentArguments(out);
```
amokiche-amd, line 636: It also better return value instead using output parametres

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -227,53 +224,23 @@ rocprofvis_result_t SystemTable::Setup(rocprofvis_dm_trace_t dm_handle, Argument
 {
     rocprofvis_result_t result = kRocProfVisResultInvalidArgument;
     
-    QueryArguments query_args;
+    TableArguments* query_args = nullptr;
```
amokiche-amd, line 227: unique_ptr looks more modern and safety here.

src/controller/src/system/rocprofvis_controller_table_system.cpp
```diff
@@ -378,36 +338,27 @@ rocprofvis_result_t SystemTable::Setup(rocprofvis_dm_trace_t dm_handle, Argument
...
+    delete query_args;
     return result;
 }
 
 rocprofvis_result_t SystemTable::ExportCSV(rocprofvis_dm_trace_t dm_handle, Arguments& args, Future* future, const char* path) const
 {
     rocprofvis_result_t result = kRocProfVisResultInvalidArgument;
     
-    QueryArguments query_args;
+    TableArguments* query_args = nullptr;
```
amokiche-amd, line 349: unique_ptr will be better

src/controller/src/rocprofvis_controller_analysis.cpp
```diff
@@ -450,8 +450,13 @@ Analysis::EventsTable::EventsTable(uint64_t id, rocprofvis_dm_event_operation_t
...
 {}
 
-rocprofvis_result_t Analysis::EventsTable::UnpackArguments(Arguments& args, QueryArguments& out) const
+rocprofvis_result_t Analysis::EventsTable::UnpackArguments(Arguments& args, TableArguments*& out) const
 {
+    if(!out)
+    {
+        out = new SystemTableArguments();
+    }
+    SystemTableArguments* sys_out = (SystemTableArguments*)out;
```
amokiche-amd, line 459: Use the dynamic_cast insted unsafe c-style cast

src/controller/src/system/rocprofvis_controller_table_system_search.h
```diff
@@ -0,0 +1,47 @@
...
+        std::vector<std::string> m_string_table_filters;
+    };
+
+    rocprofvis_result_t UnpackArguments(Arguments& args, TableArguments*& out) const override;
+    void                GetCurrentArguments(TableArguments*& out) const override;
+    void                SetCurrentArguments(TableArguments& in) override;
+
+    virtual bool                   ArgumentsChanged(SystemTableArguments& in) const;
+    virtual rocprofvis_result_t    UnpackUseCase(Arguments& args, rocprofvis_dm_table_use_case_enum_t& out) const;
+    virtual rocprofvis_dm_result_t BuildQuery(rocprofvis_dm_database_t db, TableArguments& args, uint64_t index, uint64_t count, bool count_only, char** out) const;
```
amokiche-amd: make it override

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -2265,6 +1999,182 @@ DataProvider::FetchTable(const TableRequestParams& table_params)
...
+    {
+        // request for item already exists
+        spdlog::debug("Request for this table, type {}, is already pending",
+                      static_cast<uint64_t>(table_params.m_table_type));
+        return false;
+    }
+}
+
+bool
+DataProvider::FetchEventSearch(const EventSearchRequestParams& table_params)
```
amokiche-amd, line 2015: This function has a lot of huge chunk of common code with FetchTrackTable. These parts should be moved to common helpers functions.

src/controller/src/system/rocprofvis_controller_table_system.cpp
drchen-amd: Conventionally controller uses C casts. Also it is only unsafe if you do something backwards like explicitly calling derives' from base or pass in some random type, so I don't think the cost is justified.

src/model/src/database/rocprofvis_db.cpp
drchen-amd: I don't think that makes a difference. Either base has this default unsupported implementation, or compute db has it. This class has existing examples of both.

src/controller/src/rocprofvis_controller_table.h
drchen-amd: Dosen't hurt and ensures cleanup order.

src/controller/src/system/rocprofvis_controller_table_system.cpp
drchen-amd: Controller conventionally uses raw pointers.

src/controller/src/system/rocprofvis_controller_table_system.cpp
drchen-amd: Controller conventionally manages allocations manually.

src/view/src/rocprofvis_data_provider.cpp
drchen-amd: Can you be more specific which huge chunk are common?

I see these are common:
1. Setting request id to the export id for export requests.
2. Checking if request already in progress.
3. Allocating controller argument,array,future.
4. Calling export api for export requests.
5. Allocating request info.

Each of these are a few lines and their callsites are scattered, so I don't think having DataProvider::SetExportRequestID, DataProvider::RequestExists, DataProvider::AllocateArgument...etc would be that helpful.

The bulk of these functions are setting the bespoke arguments/request info for each path.

src/controller/src/system/rocprofvis_controller_table_system.cpp
drchen-amd: Can you elaborate what the function signature and implementation at each level would look like?

src/controller/src/system/rocprofvis_controller_table_system.cpp
tomk-amd: Yeah, let's keep to the style the controller currently is using.  (If / when we refactor this module we can change the approach everywhere).

src/controller/src/system/rocprofvis_controller_table_system.cpp
tomk-amd: I think it is ok to leave the signature as currently implemented.  I believe this falls into one of the "Exception" cases in the guidline doc anyway.

src/model/src/database/rocprofvis_db.cpp
tomk-amd: I think the point is to return kRocProfVisDmResultNotSupported;

This is the pattern used in the model, derived don't implement and instead rely on base to return the error.

src/controller/src/system/rocprofvis_controller_table_system.cpp
amokiche-amd: Just because the controller previously used raw pointers doesn't mean we're forced to follow suit. We have no external restrictions on this, and we can easily change it to use safer and more modern language features. The problem with c-style casting isn't how it works now, but the potential problems it poses when it's changed. We don't lose anything by using modern casts, but we gain more security because the code more clearly communicates the intended cast's intentions.
If you definitely sure about here it always will be a cast from parent to child the static_cast will be better option for it. You avoid the vtable search and show your intentions.

src/model/src/database/rocprofvis_db.cpp
amokiche-amd: But the base class already has pure virtual functions, so it can't be instantiated. And the child function will always be called using the pointer. The only way this function can be called is by calling it directly insyde the child function, but I don't see any direct calls to it in the entire code. And even if there were, I think it would be easier to simply return kRocProfVisDmResultNotSupported in the child function than to call the parent function. There's no way for this function to be called in our code, so I think it's better to make it pure virtual.

src/controller/src/rocprofvis_controller_table.h
amokiche-amd: Adding a virtual destructor increases the size of the structure by the size of the virtual table pointer. It also removes the structure from the primitive structure category, which may prevent the compiler from making certain optimizations. If this structure isn't intended as a base for another, a virtual destructor is only a nuisance and confuses the reader of the code.
And the presence of a virtual destructor does not add any order.

src/controller/src/system/rocprofvis_controller_table_system.cpp
amokiche-amd: This is a great opportunity to start reworking this, as history has shown that smart pointers don't lose efficiency while providing greater security. Again, I see no reason not to start using them, especially considering this is a refactoring pull request.

src/controller/src/system/rocprofvis_controller_table_system.cpp
amokiche-amd: This is clearly not a critical thing, so I will not insist on it, but the function looks like a canonical factory.

src/controller/src/system/rocprofvis_controller_table_system.cpp
amokiche-amd: Again, I think this is a great place to start refactoring, but I'm okay with doing this in a separate PR.

src/view/src/rocprofvis_data_provider.cpp
amokiche-amd: Provider-state validation
Export request-ID selection
Argument allocation
Future and array allocation
Async fetch/export submission
RequestInfo construction
Request insertion

This parts are common and it can be moved to separate functions, that make code more readable and the body of function less. Even helper function uses only in one place it good patern to move this functionality to separate function, to avoid large functions with hundreds of lines.

src/controller/src/rocprofvis_controller_table.h
drchen-amd: What do you mean "if this structure isn't intended as a base for another"? It is already sub-classed by systemtable and eventsearch variants. They even have vector members.

src/view/src/rocprofvis_data_provider.cpp
drchen-amd: - Except the export, those are common to all requests and have little to do with table API. 
- Argument, Future, array allocation are single line function calls.  
- Request info allocation will have a huge amount of function parameters if you wish to factor into a common function because all requests have their own unique parameters. Same with async fetch call to a lesser extend as requests are services by different controller API .

src/controller/src/rocprofvis_controller_table.h
amokiche-amd: No it not. A struct declared inside another class is a nested type. It does not become part of the inheritance hierarchy. The fact you declare it inside the class influence only on scope of view of this struct.

drchen-amd:
```
A review comment said the tableargument does not need virtual deconstructor because it is not part of inherhitence hierarchy. Is this true? : "No it not.
  A struct declared inside another class is a nested type. It does not become part of the inheritance hierarchy. The fact you declare it inside the class
  influence only on scope of view of this struct."
```

```
● Let me check where those derived structs are declared:

  Read 1 file (ctrl+o to expand)

● The reviewer's fact is true but their conclusion is false — they're reasoning about the wrong hierarchy.

  What's true

  "A struct declared inside another class is a nested type; nesting only affects scope, not inheritance." Correct. Declaring TableArguments inside Table
  does not make TableArguments part of Table's inheritance hierarchy. TableArguments is not a base or derived class of Table. Nesting == scope, nothing
  more. No argument there.

  Why it's irrelevant

  There are two independent hierarchies here, and the reviewer is talking about the one that doesn't matter:

  Hierarchy 1 — the enclosing classes (what the reviewer is thinking about):
  Table ← SystemTable ← EventSearchTable
  TableArguments is not in this one. Agreed.

  Hierarchy 2 — the argument structs themselves (what actually matters):
  TableArguments ← SystemTableArguments ← EventSearchTableArguments
  This hierarchy is real. It's created by explicit inheritance clauses, which you can see directly:
  - struct SystemTableArguments : TableArguments (rocprofvis_controller_table_system.h:39)
  - struct EventSearchTableArguments : SystemTableArguments (rocprofvis_controller_table_system_search.h:28)

  Nesting and inheritance are orthogonal. A nested struct can absolutely be a base class. SystemTableArguments is both nested (inside SystemTable) and a
  derived class of TableArguments. The nesting doesn't cancel the : TableArguments. So the premise "TableArguments is not part of an inheritance hierarchy"
  is simply wrong — TableArguments is the base of hierarchy 2.

  The test that actually decides it

  The virtual-destructor rule doesn't care about nesting at all. It cares about one thing: is a derived object ever deleted through a base pointer? Here,
  yes:

  - new SystemTableArguments() / new EventSearchTableArguments() are allocated behind a TableArguments*,
  - then destroyed via delete fetch_args; and delete query_args; (rocprofvis_controller_table_system.cpp:165, 341, 382), where the static type is
  TableArguments*.

  delete on a base pointer whose class has a non-virtual destructor is undefined behavior when the dynamic type is derived — and since the derived structs
  add members (std::string, std::vector), it would also leak/mis-destruct those in practice. The virtual ~TableArguments() = default; is exactly what makes
  those deletes well-defined.

  Bottom line

  The reviewer's statement about nesting is correct but is a non-sequitur. The virtual destructor is required — not because of where the struct is declared,
  but because derived …TableArguments objects are deleted through a TableArguments*. Keep it.
```

amokiche-amd: TableArguments has the child struct? ... So we've been talking about different things all this time, and I didn't see that you inherited SystemTableArguments from it. Then yes, that clears up all the questions :) You were right here

tomk-amd (approved):

PR #961, opened 2026-07-28
refactor(optiq): consume track-based profiler-hub reader API (opaque ids, unified detail)

avansick-amd:
## Motivation

Migrates roc-optiq onto the refactored profiler-hub reader API. The reader now
exposes a track-based surface with opaque ids and typed accessors (see
ROCm/rocm-systems#9366); this branch updates Optiq's data model, controller, and
session-cache layer to consume that surface directly, so Optiq no longer reaches
around the reader for detail data.

## Technical Details

- **Opaque ids end-to-end:** Optiq consumes `event_id_t`, `flow_id_t`, and
  `track_id_t` as opaque types. The reader remains the sole gateway; Optiq only
  reads the raw `track_id_t` value where it must serialize `reader_track_id` to
  the session-cache DB (integer column).
- **Unified detail path:** event/counter detail is sourced through the reader's
  unified `get_event_detail` / `get_event_info` accessor rather than Optiq's
  former direct-SQL detail queries.
- **Track model alignment:** lane + nesting_model and typed/directed flows from
  the reader are wired through the model -> controller -> view layers.
- **Name-fidelity follow-through:** identifiers updated to match the reader's
  spec-aligned names.

History note: this branch is 18 commits cut from `dev` (an ancestor of the
default `main` branch); `main` is 96 commits ahead. It will be rebased onto fresh
`origin/main` and curated before this draft is marked Ready for review.

## Test Plan

Datamodel target build + Optiq unit suite; interactive GUI panel-fidelity check.

## Test Result

Datamodel target builds green. GUI panel-fidelity human gate still pending (called
out below).

## Reviewer Notes

- Draft for visibility -- paired with the profiler-hub reader-API PR
  (ROCm/rocm-systems#9366). Review the reader PR alongside this one.
- **Pending human gate:** interactive GUI panel-fidelity check has not yet been
  performed; please do not treat rendering parity as verified.
- Rebase onto latest `origin/main` + history curation still owed before
  Ready-for-review.


vstempen:
Could you please implement a method to detect if file is supported by profiler hub.

vstempen:
I don't think this is proper way of listing UUIDs:
    auto uuid_query_executor = create_read_statement_executor<uuid_result>(
        "SELECT DISTINCT replace(name, rtrim(name, replace(name, '_', '')), '') "
        "AS guid "
        "FROM sqlite_master WHERE type='table' AND name LIKE 'rocpd_%';",
&uuid_result::uuid);
First of all the query only returns last portion of uuid. So, if say table name is rocpd_metadata_0000fc9e_5637_7637_9e70_5e51139613a3, you will get  5e51139613a3.
I think proper query would be : 
select replace(guid,'-','_') from rocpd_info_node;

vstempen:
Please add LOG_INFO for every query execution and a way to turn it on and off.

vstempen:
Profiler hub should handle exceptions and have an enumeration of execution states. Interface methods should return the state values or throw exceptions describing a state.

PR #962, opened 2026-07-28
Render Memory Chart from JSON description document

dhingora-amd:
## Motivation

The compute kernel-details Memory Chart was a static, hardcoded diagram that couldn't adapt to different GPU architectures or traces. This makes it data-driven: the memory hierarchy is described as JSON stored per-workload in the trace database, so the chart reflects each workload and can evolve without code changes.

## Technical Details

- **Relational layout model** (`rocprofvis_memory_chart_model.{h,cpp}`): a set of blocks (nodes) and arrows (edges) parsed from JSON. Blocks nest via `children` (container boxes); metrics are referenced by their full dotted id `category.table.entry`; a semantic `category` drives accent colors. No coordinates in JSON — columns, block sizing, and arrow routing are computed at runtime.
- **Layout from the database**: new optional `compute_workload.memory_chart_extdata` column, detected via PRAGMA and plumbed model → controller C ABI (`kRPVControllerWorkloadMemoryChartLayout`) → data provider → view, with an embedded default (`rocprofvis_memory_chart_default_layout.h`) as fallback.
- **Renderer** (`rocprofvis_compute_memory_chart.{cpp,h}`): resolves the selected kernel's metric values and draws the chart. Skip-arrow routing packs "highway" lanes (horizontally disjoint arrows share a lane; shorter spans nest inside longer ones), spreads each block's bottom connectors symmetrically about its centre, and draws all lines before labels so no line is painted over a label.
- Docs updated (`.agents/AGENTS.md`)


dhingora-amd:
[memory_chart.schema.json](https://github.com/user-attachments/files/30468623/memory_chart.schema.json)
Above is the proposed schema.

src/view/src/compute/rocprofvis_memory_chart_default_layout.h
```diff
@@ -0,0 +1,103 @@
...
+namespace RocProfVis
+{
+namespace View
+{
+
+inline constexpr const char* kDefaultMemoryChartLayout = R"JSON(
+{
+    "version": 2,
+    "metric_source": { "category_id": 3, "table_id": 1 },
+    "blocks": [
```
tomk-amd: The version should start from 1 not 2.

I think the metric_source field should be removed.   Metrics can come from more than one source, ex: 3.1.2, 3.3.2, etc...

src/view/src/compute/rocprofvis_memory_chart_model.h
```diff
@@ -0,0 +1,113 @@
...
+
+// A reference to a metric. The data model addresses metrics by numeric entry id;
+// a name may be used instead for authoring convenience. Exactly one of the two
+// forms is populated when `valid` is true.
+struct MemChartMetricRef
+{
+    bool        valid = false;
+    bool        by_id = false;
+    uint32_t    id    = 0;
+    std::string name;
```
tomk-amd: if the metric_source field is removed than id and by_id should be removed as well.  `by_id` and `id` don't seem to be used anyway..

dhingora-amd:
[memory_chart.schema.json](https://github.com/user-attachments/files/30513151/memory_chart.schema.json)
Updated Schema.

src/model/src/database/rocprofvis_db_compute.cpp
```diff
@@ -1099,6 +1104,9 @@ void ComputeQueryFactory::ParseMetricParam(std::string metric_str, uint32_t work
 		{			
 			if (kRocProfVisDmResultSuccess != ExecuteSQLQuery(future, &tmp_db_instance, "SELECT * FROM compute_metadata;", &CallbackParseMetadata)) break;
 			m_query_factory.SetVersion(m_db_version.c_str());
+			// Best-effort detection of the optional memory_chart_extdata column so
+			// the workload query includes it only when present.
+			ExecuteSQLQuery(future, &tmp_db_instance, "SELECT name FROM pragma_table_info('compute_workload') WHERE name = 'memory_chart_extdata';", &CallbackDetectMemoryChartColumn);
```
drchen-amd, line 1109: Does this mean this feature does not come with schema bump?

I think preferably the schema gets bumped, and above that version we can expect this column, in which case we would not need a bespoke query for detection.

tomk-amd: No schema bump for now (unknown what the number will be when this is ready,  the compute side isn't ready to implement the JSON blob in the DB yet).  But we are moving ahead with hard coded JSON template files as this will fix the memory chart for different architectures for the upcoming release at least.

drchen-amd (commented):
Can you merge in latest main? Metrics for newer db versions are mapped incorrectly and there is a change that should address it.

A couple of issues from testing:
- Metric tooltips get cut off a few ticks above minimum font size.
- % metrics like hit rate and utilization now display without units where old version had them. 
image

src/view/src/compute/rocprofvis_memory_chart_model.cpp
```diff
@@ -0,0 +1,257 @@
+// Copyright Advanced Micro Devices, Inc.
```
drchen-amd: This file and its header should be moved in the model subdirectory:
\src\view\src\model\compute\

src/view/src/model/compute/rocprofvis_memory_chart_model.cpp
```diff
@@ -0,0 +1,258 @@
+// Copyright Advanced Micro Devices, Inc.
```
drchen-amd, line 1: This file and its header should be moved in the model subdirectory:
\src\view\src\model\compute\

src/view/src/compute/rocprofvis_compute_memory_chart.cpp
```diff
@@ -362,912 +347,1193 @@ DrawLegend(ImDrawList* draw_list, ImVec2 origin, float y)
...
-        std::max(canvas_w, ImGui::GetContentRegionAvail().x + ImGui::GetScrollX());
+    uint32_t workload_id = m_compute_selection->GetSelectedWorkload();
+    uint32_t kernel_id   = m_compute_selection->GetSelectedKernel();
 
-    DrawChartBackdrop(draw_list, window_position, backdrop_w, canvas_h);
+    // Metrics may span multiple categories/tables (e.g. 3.1.x and 3.3.x). Fetch
+    // each category the layout references, whole (all sub-tables), so every
+    // referenced metric resolves.
+    std::set<uint32_t> categories;
+    CollectCategories(m_layout.blocks, categories);
```
drchen-amd: The controller API supports grabbing individual tables and individual metrics, so this seems overkill for how sparse these new charts are. Ex for gfx940, you need four metrics from 17.2, but the entire 17 is fetched. 

I think a good balance would be to fetch by individual tables instead of the entire category.

src/view/src/model/compute/rocprofvis_memory_chart_model.cpp
```diff
@@ -0,0 +1,258 @@
...
+        }
+        MemChartBlock* found = FindInBlocks(block.children, id);
+        if(found)
+        {
+            return found;
+        }
+    }
+    return nullptr;
+}
+}  // namespace
```
drchen-amd, line 243: This id scan happens 6-10 times per arrow per render cycle, so low hanging fruit would be an `id -> block` map. 

I don't think this mapping is dynamic (arrows always point to same blocks for a given gfx_arch)? If they are fixed, runtime lookup cook be replaced with each arrow carrying properties for which blocks they link.

src/view/src/compute/rocprofvis_compute_memory_chart.cpp
```diff
@@ -362,912 +347,1193 @@ DrawLegend(ImDrawList* draw_list, ImVec2 origin, float y)
...
+    // Leaf: size from its metric rows.
+    float width = ImGui::CalcTextSize(block.title.c_str()).x;
+    for(const MemChartContentItem& item : block.content)
     {
-        ImGui::SetTooltip(
-            "%s",
-            FormatMetricValueRaw(m_metric_ptrs[metric_id]->values.begin()->second)
-                .c_str());
+        std::string label = MetricLabel(item.metric, item.title);
+        std::string value = MetricValueText(item.metric, true, item.unit);
```
drchen-amd: This handles conversion to string and formatting for the metric value but is called once here in `Measure `and once again in `Draw` in a render cycle. Once you fetched your metric, the value cannot change, so this is a good candidate for caching.

src/view/src/model/compute/rocprofvis_compute_model_types.h
```diff
@@ -176,6 +176,7 @@ struct WorkloadInfo
     std::string                              name;
     std::vector<std::vector<std::string>>    system_info;
     std::vector<std::vector<std::string>>    profiling_config;
+    std::string                              memory_chart_layout;  // Raw JSON layout from the DB (empty if absent).
```
drchen-amd: If instead of raw json, you stored your parsed `MemChartLayout` here, you wouldn't need to re-parse json each time workload/kernel changes.

drchen-amd (commented):
This seems to lack a "render friendly" memory chart model, and as a result there is a fair amount of redundant processing to make it render friendly every frame. For example:

- `ComputeMemoryChartView::MetricLabel` - Returns a passed in overridden name if exists, otherwise metric name if metric exists.

- `ComputeMemoryChartView::MetricValueText` - Converts metric value to str, appends unit.
- Arrows and blocks searching for for which blocks they connect to.

There is a `rocprofvis_memory_chart_model`, but it is a collecting of helper functions and does not hold any data. This could be expanded to provide a "friendly" data representation. Rough idea being that it would store a bunch of these:

```
struct Metric
{
  int category, table, entry
  string name
  string value
}

struct Block
{
  int id,
  string name,
  vector<Metric> metrics
  Imvec2 x,ypos
  Imvec2 length,width
}

struct Arrow
{
  int id,
  int name,
  Metric metric
  Block* source
  Block* target
}
```

Then your `ComputeMemoryChartView` would update subsections of this only when needed:

- On workload change, your entire layout changes, so you would create fresh Blocks and Arrows.
- On kernel change, you go through your Arrows and Blocks, and update only the metric related data.

The sizing and positioning updates need to remain in render, but everything else would be ready as is for render to consume.

tomk-amd (approved):

tomk-amd:
Merging - but this should be addressed soon if real:

> AI Review

The one real gap bug. Skip arrows (column distance greater than 1) mark every intervening gap as arrow-bearing, but they are not drawn through those corridors — they go under the blocks on the highway, or up the left margin. A skip from L2 to HBM would re-open the MALL/UMC gaps this change is trying to close. default.json already has a skip (13 → 1); those gaps stay wide today only because adjacent arrows also cross them.

PR #963, opened 2026-07-28
Compare Improvements

Sajeeth-Wimalasuriyan:
Analysis panel did not have a meaningful way to compare but now this has been added across all tabs in the panel where applicable. 
image
image


src/view/src/rocprofvis_compare_panes.h
```diff
@@ -0,0 +1,74 @@
...
+class HSplitContainer;
+class SettingsManager;
+class TraceDataModel;
+struct CompareSourceInfo;
+
+// Compare sources are ordered as the project opened them, which is also what
+// TrackInfo::file_id counts.
+constexpr size_t COMPARE_SOURCE_A     = 0;
+constexpr size_t COMPARE_SOURCE_B     = 1;
+constexpr size_t COMPARE_SOURCE_COUNT = 2;
```
amokiche-amd, line 29: you ara using this check in the Project::OpenCompare()
if(file_paths.size() < 2 || m_view)
    {
        return result;
    }
Do we have a plans to compare more than two traces in the future?
Becouse now this coment and chekc litle bit confucing

Image
If we want to compare only two traces magic valuw "2" should be replaced by the constant,

src/view/src/rocprofvis_multi_track_table.cpp
```diff
@@ -33,19 +38,183 @@ MultiTrackTable::MultiTrackTable(DataProvider& dp, TableType table_type,
...
+MultiTrackTable::RenderSharedFilterControls()
+{
+    // Same shape as the per table form: a label column, the group by combo, the
+    // apply button on the right, then a filter input filling the row.
+    const ImGuiStyle& style     = ImGui::GetStyle();
+    ImFont*           icon_font = m_settings.GetFontManager().GetFont(FontType::kIcon);
+    const bool        grouping  = !m_pending_filter_options.group_by.empty();
+
+    const float label_col =
+        ImGui::CalcTextSize("Group by").x + style.ItemSpacing.x * SHARED_LABEL_GAP;
```
amokiche-amd: Make the text constexpr variable insted to use it twice

src/view/src/rocprofvis_multi_track_table.cpp
```diff
@@ -33,19 +38,183 @@ MultiTrackTable::MultiTrackTable(DataProvider& dp, TableType table_type,
...
+    const float apply_width =
+        ImGui::CalcTextSize(SHARED_APPLY_LABEL).x + style.FramePadding.x * 2.0f;
+
+    ImGui::Spacing();
+
+    ImGui::AlignTextToFramePadding();
+    ImGui::TextUnformatted("Group by");
```
amokiche-amd: We use for compare view "Group by", but in single trace view we are using label "Agregate". Different labels for the same action can confuse the user. Better to chose one for the both views (I prefer the "Group by" but you can chose anything)

src/view/src/rocprofvis_compare_panes.h
```diff
@@ -0,0 +1,74 @@
...
+
+// Matches CompareSourceInfo::id, for labels built before the sources are known.
+constexpr const char* COMPARE_SOURCE_LABEL[COMPARE_SOURCE_COUNT] = { "A", "B" };
```
amokiche-amd, line 32: nit: I think that inline constexpr std::array looks more modern, but I don't insist on it.

amokiche-amd (commented):
All looks good but it has a small bug. If you open the compare view, than select something only from the B trace. Tan press Group by: you will see only the "None" column

Image

If any A element was selected befor it will be ok, also if after reproducing it you select something from the A trace and unselect it will be ok.
If only B has selected tracks, A has no schema and the shared group-by selector contains only “None,” despite B having valid columns. The shared filter state should use the first available schema or be independent of either pane.

tomk-amd:
Some issues that will need to be addressed in a follow up PR.
[COMPARE_REVIEW_FOLLOWUP.md](https://github.com/user-attachments/files/32028564/COMPARE_REVIEW_FOLLOWUP.md)

tomk-amd (approved):

PR #967, opened 2026-07-29
View: Add "Show Hidden Tracks" to the track context menu

dhingora-amd:
## Motivation

Hidden timeline tracks could only be restored from the topology side bar. If
that panel is closed, there was no way to bring them back. This adds that
option to a track's right-click menu.

## Technical Details

- New **Show Hidden Tracks** entry in the track context menu, shown only when
  tracks are hidden.
- Lists hidden tracks grouped by category (queues, streams, threads,
  counters), each with **Show All Hidden**, plus a top-level **Show All Hidden
  Tracks**.
- Revealing a track fires `kTrackVisibilityChanged`, the same path the side bar
  uses. Category submenus are height-capped and scroll for long lists.
- Menu fix: the icon column is now a fixed width, so labels align regardless of
  the leading icon's glyph width (e.g. eye vs. gear). Affects all icon menus.

### DIFF
--- a/src/view/src/rocprofvis_timeline_view.cpp
+++ b/src/view/src/rocprofvis_timeline_view.cpp
@@ -37,6 +37,7 @@ constexpr float    SCROLL_SPEED                  = 100.0f;
 constexpr uint64_t DEFAULT_LOADING_TIMER         = 150;  // milliseconds
 constexpr float    ARTIFICIAL_SCROLLBAR_HEIGHT   = 18.0f;
 constexpr float    SIDEBAR_SPLITTER_WIDTH        = 5.0f;
+constexpr const char* HIDDEN_TRACKS_MENU_POPUP_NAME = "HiddenTracksMenu";
 // Build a text block mirroring the on-hover tooltip (name, timing, and id)
 // for the clipboard.
 static std::string
@@ -1754,6 +1755,8 @@ TimelineView::RenderGraphView()
         RenderTrack(index, request_data, window_flags, container_size);
     }
 
+    RenderEmptyTrackAreaMenu();
+
     TrackItem::SetSidebarSize(m_sidebar_size);
     ImGui::EndChild();
     ImGui::PopStyleColor();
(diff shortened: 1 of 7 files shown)


src/view/src/rocprofvis_track_item.cpp
```diff
@@ -557,6 +557,16 @@ TrackItem::RenderMetaArea()
             m_timeline_track_options.RenderContextMenu();
             ImGui::EndMenu();
         }
+        // Restore hidden tracks without needing the topology side bar. The entry
+        // is only shown when something is actually hidden.
+        if(m_timeline_track_options.HasHiddenTracks())
+        {
+            if(IconBeginMenu(ICON_EYE, "Show Hidden Tracks"))
+            {
+                m_timeline_track_options.RenderHiddenTracksSubmenu();
```
drchen-amd, line 566: This use case should not be initiated from a track item because it needs to account for scenario when all track items are hidden. When all track items are hidden, current implementation is not accessible.

drchen-amd: On second thought; it can be accessed from a track item, but it also needs to be accessible when all track items are hidden.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -1805,6 +1808,66 @@ TimelineView::IsRequestDataNeeded()
...
+{
+    if(!m_tracks) return false;
+
+    for(const TrackItem* track : *m_tracks)
+    {
+        if(track && track->IsDisplayed()) return true;
+    }
+    return false;
+}
+
```
tomk-amd: This gets called every frame, is there a better way to get this state instead of traversing the entire list each time?

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -1805,6 +1808,54 @@ TimelineView::IsRequestDataNeeded()
     return request_data;
 }
 
+void
+TimelineView::RenderEmptyTrackAreaMenu()
+{
+    if(!m_track_options_context_menu || !m_track_options_context_menu->HasHiddenTracks())
+    {
+        return;
+    }
```
tomk-amd, line 1820: This also will loop through all of the tracks every frame.

HasHiddenTracks() also iterates through all the tracks.

tomk-amd (approved):

PR #968, opened 2026-07-29
Do not load workload speed of light for schema < 1.3.0

drchen-amd:
Do not attempt to query `compute_workload_metric_view` when opening compute schema < 1.3.0 files. 

Compute query builders already perform schema version checks, but they only return the query string. Controller is left with binary query empty/not empty and other context is lost. Instead, have query builders return result code. 

Note:
This is not the minimal change required to avoid querying `compute_workload_metric_view` for older schemas. If we want to stick with minimal change, I can switch to that. Otherwise, I think having query builders give more finer results codes to controller could be useful should they need to conditionally skip queries, such as when metric definitions...etc are missing. In these cases, query builder could return not supported/not loaded, and controller could use that as signal to skip loading particular table instead of trying and showing error.

### DIFF
--- a/src/model/src/database/rocprofvis_db_compute.h
+++ b/src/model/src/database/rocprofvis_db_compute.h
@@ -24,26 +24,26 @@ namespace DataModel
     {
     public:
         ComputeQueryFactory(ComputeDatabase* db) : m_db(db) {}
-        rocprofvis_dm_result_t GetComputeListOfWorkloads(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
-        rocprofvis_dm_result_t GetComputeWorkloadRooflineCeiling(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
-        rocprofvis_dm_result_t GetComputeWorkloadTopKernels(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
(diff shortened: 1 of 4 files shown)


src/model/src/database/rocprofvis_db_compute.h
```diff
@@ -24,26 +24,26 @@ namespace DataModel
...
-        std::string GetComputeKernelIsaToIsaDeps(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params);
-        std::string GetComputeKernelIsaLines(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params);
-        std::string GetComputeKernelIsaToSourceDeps(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params);
-        std::string GetComputeKernelSamplingStates(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params);
-        std::string GetComputeKernelSamplingStateReasonCounts(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params);
+        rocprofvis_dm_result_t GetComputeListOfWorkloads(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
+        rocprofvis_dm_result_t GetComputeWorkloadRooflineCeiling(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
+        rocprofvis_dm_result_t GetComputeWorkloadTopKernels(rocprofvis_db_num_of_params_t num, rocprofvis_db_compute_params_t params, rocprofvis_dm_string_t& query);
```
tomk-amd: nitpik: maybe rename the `query` parameter to `query_out` ?

tomk-amd (approved):

drchen-amd:
(Updated DATABASE.md)

PR #970, opened 2026-07-30
RPD Table & Extdata Update

drchen-amd:
1. Correct naming disagreements between flame chart and tables for Api events:
    - Previous: flame chart name = `apiName`, table name column = `args`, event details name = `args`
    - New: flame chart name = `apiName`, table name column = `apiName`, event details name = `apiName`
2.  Replaced table `category` column for api events with `arguments` column. 
    - op events have a `opType` field that is mapped to category, api events do not have this field. They instead have `args`.
4. Fixed incorrect table `category` values for op events.
    - `CATEGORY_REFERENCE_RPD` should be `qword` instead of `word`.

This addresses api events unexpectedly having blank names, but it is still expected that there are op events with blank names:
image


tomk-amd (approved):
Looks good but maybe @vstempen should give it blessing too

vstempen (approved):

PR #975, opened 2026-07-31
Correct missing flame labels, track minimum height

drchen-amd:
- Reconcile flame event box and label heights to use text height instead of literal font size.
  - Similarly, update track minimum height. Seems like the intent was to ensure pills always had room to render, but this was not working for the same reason.

Note: It cannot be the reverse (everybody uses literal font size). In that case label would render, but its size may overflow/be too small compared to the event box.


Sajeeth-Wimalasuriyan (approved):
This fix is better than the one I made. I have tested the fix it is all good.

PR #981, opened 2026-08-05
Refactor controller_track and system_trace

amokiche-amd:
## Motivation

Refactor controller Track and System_trace files to replace of usage default getters to direct getters and setters.
Refactor giant functions to increase code readability.
Small minor refactors.

## Submission Checklist


src/controller/src/system/rocprofvis_controller_track.cpp
```diff
@@ -519,31 +657,31 @@ rocprofvis_result_t Track::GetUInt64(rocprofvis_property_t property, uint64_t in
             }
             case kRPVControllerTrackNumberOfEntries:
             {
-                *value = m_num_entries;
+                *value = m_bounds.num_entries;
```
tomk-amd: use GetNumberOfEntries() to prevent risk of ABI getters drifting from C++ interface

src/controller/src/system/rocprofvis_controller_track.cpp
```diff
@@ -519,31 +657,31 @@ rocprofvis_result_t Track::GetUInt64(rocprofvis_property_t property, uint64_t in
             }
             case kRPVControllerTrackNumberOfEntries:
```
tomk-amd, line 879: use GetTrackType() instead of `m_type` to prevent risk of ABI getters drifting from C++ interface

drchen-amd (commented):
I don't mind this new approach of using more C++ features and directly calling functions on the objects internally instead of enum. I also do not mind sticking to C style and using enums internally. 

I think new approach will lead to some bloating as for each exposed field you end up with an enum version and a direct call version.

If we proceed, this will need more hardening because I only tested for a few minutes and noticed:
-  Queue utilization and counter track stats are broken.
-  Panning and zooming will debug assert: Image

amokiche-amd:
> * Queue utilization and counter track stats are broken.
> * Panning and zooming will debug assert: 

Nice catch, thanx queue and counter utilization fixed.
All fixed.

amokiche-amd:
I removed the changes with replacement raw pointers to the unique_ptr, because it has unobvious dependency and it work for other separate PR.

src/controller/src/system/rocprofvis_controller_topology.cpp
tomk-amd, line 52: set `child_track` to nullptr here - GetObject doesn't guarantee it will set it to null on failure ... existing issue I believe, but since we're cleaning up :) might as well fix.

src/controller/src/system/rocprofvis_controller_track.cpp
```diff
@@ -999,8 +1150,8 @@ rocprofvis_result_t Track::SetObject(rocprofvis_property_t property, uint64_t in
                 ThreadRef ref(value);
                 if(ref.IsValid())
                 {
-                    m_thread = ref.Get();
-                    result = kRocProfVisResultSuccess;
+                    m_topology_links.thread = ref.Get();
+                    result                  = kRocProfVisResultSuccess;
                 }
                 break;
```
tomk-amd: For streams, queues, threads, etc.. use the setters?

                    SetThread(ref.Get());
                    result = kRocProfVisResultSuccess;

tomk-amd (approved):

PR #982, opened 2026-08-05
View: Add sortable timeline tracks (topology/default/custom)

dhingora-amd:
## Motivation

Give users a quick way to organize timeline tracks instead of only manual
drag-reorder. Also decouples reordering from the controller — its graph order
is presentation-only (fetch is keyed by track id), so the round-trip was
unnecessary.

## Technical Details

- Right-click the track-list header → *Sort tracks by*: **Topology** (sidebar
  order), **Default** (load order), **Custom** (remembered manual order; drag
  auto-switches to it).
- Reordering is now view-only: updates `TrackInfo::index` + fires
  `kTrackMetadataChanged`; removed the controller `SetGraphIndex` sync.
- Topology order cached/deduped in `TrackTopology` (computed once per tree build).
- Persists `sort_mode` + `custom_order` per project (validated before applying).
- Menu matches the other context menus (`IconMenuItem` + base-style padding).

### DIFF
--- a/src/view/src/rocprofvis_data_provider.cpp
+++ b/src/view/src/rocprofvis_data_provider.cpp
@@ -286,7 +286,7 @@ DataProvider::SetRequestProgressUpdateCallback(
 }
 
 bool
-DataProvider::SetTrackDisplayOrder(const std::vector<uint64_t>& ordered_track_ids)
+DataProvider::SetTrackIndex(const std::vector<uint64_t>& ordered_track_ids)
 {
     if(m_state != ProviderState::kReady)
     {
@@ -296,10 +296,10 @@ DataProvider::SetTrackDisplayOrder(const std::vector<uint64_t>& ordered_track_id
     TimelineModel&                           tlm      = m_model.GetTimeline();
     std::unordered_map<uint64_t, TrackInfo>& metadata = tlm.GetMutableTrackMetadata();
 
-    // Require a full permutation so no two tracks share a display index.
+    // Require a full permutation so no two tracks share an index.
     if(ordered_track_ids.size() != metadata.size())
     {
-        spdlog::warn("SetTrackDisplayOrder ignored: expected {} track ids, got {}",
+        spdlog::warn("SetTrackIndex ignored: expected {} track ids, got {}",
                      metadata.size(), ordered_track_ids.size());
         return false;
     }
@@ -310,7 +310,7 @@ DataProvider::SetTrackDisplayOrder(const std::vector<uint64_t>& ordered_track_id
         auto it = metadata.find(track_id);
         if(it == metadata.end())
         {
-            spdlog::warn("SetTrackDisplayOrder ignored: unknown track id {}", track_id);
+            spdlog::warn("SetTrackIndex ignored: unknown track id {}", track_id);
             return false;
         }
         it->second.index = index++;
--- a/src/view/src/rocprofvis_data_provider.cpp
+++ b/src/view/src/rocprofvis_data_provider.cpp
@@ -304,7 +304,11 @@ DataProvider::SetTrackIndex(const std::vector<uint64_t>& ordered_track_ids)
         return false;
     }
 
-    uint64_t index = 0;
+    // Resolve every id up front so a bad order can't leave the metadata
+    // half-reindexed (some tracks new indexes, the rest stale). We only mutate
+    // once the whole order is confirmed valid.
+    std::vector<TrackInfo*> ordered_metadata;
+    ordered_metadata.reserve(ordered_track_ids.size());
     for(uint64_t track_id : ordered_track_ids)
     {
         auto it = metadata.find(track_id);
@@ -313,7 +317,13 @@ DataProvider::SetTrackIndex(const std::vector<uint64_t>& ordered_track_ids)
             spdlog::warn("SetTrackIndex ignored: unknown track id {}", track_id);
             return false;
         }
-        it->second.index = index++;
+        ordered_metadata.push_back(&it->second);
+    }
+
+    uint64_t index = 0;
+    for(TrackInfo* info : ordered_metadata)
+    {
+        info->index = index++;
     }
 
     if(m_track_metadata_changed_callback)
--- a/src/view/src/rocprofvis_data_provider.cpp
+++ b/src/view/src/rocprofvis_data_provider.cpp
@@ -286,47 +286,41 @@ DataProvider::SetRequestProgressUpdateCallback(
 }
 
 bool
-DataProvider::SetGraphIndex(uint64_t track_id, uint64_t index)
+DataProvider::SetTrackDisplayOrder(const std::vector<uint64_t>& ordered_track_ids)
 {
-    rocprofvis_result_t result = kRocProfVisResultUnknownError;
-    TimelineModel& tlm = m_model.GetTimeline();
-    uint64_t num_graphs = tlm.GetTrackCount();
-
-    if(m_state == ProviderState::kReady)
+    if(m_state != ProviderState::kReady)
     {
-        ROCPROFVIS_ASSERT(index < num_graphs);
-        const TrackInfo* metadata = tlm.GetTrack(track_id);
-        ROCPROFVIS_ASSERT(metadata);
-        result = rocprofvis_controller_set_object(m_trace_timeline,
-                                                  kRPVControllerTimelineGraphIndexed,
-                                                  index, metadata->graph_handle);
-        if(result == kRocProfVisResultSuccess)
-        {
-            rocprofvis_handle_t* graph = nullptr;
-            for(int i = 0; i < num_graphs; i++)
-            {
-                result = rocprofvis_controller_get_object(
-                    m_trace_timeline, kRPVControllerTimelineGraphIndexed, i, &graph);
-                if(result == kRocProfVisResultSuccess && graph)
-                {
-                    uint64_t id = 0;
-                    result      = rocprofvis_controller_get_uint64(
-                        graph, kRPVControllerGraphId, 0, &id);
-                    if(result == kRocProfVisResultSuccess)
-                    {
-                        metadata = tlm.GetTrack(id);
-                        ROCPROFVIS_ASSERT(metadata && metadata->graph_handle == graph);
-                        tlm.GetMutableTrackMetadata()[id].index = i;
-                    }
-                }
-            }
-            if(m_track_metadata_changed_callback)
-            {
-                m_track_metadata_changed_callback(m_model.GetTraceFilePath());
-            }
-        }
+        return false;
     }
-    return (result == kRocProfVisResultSuccess);
+
+    TimelineModel&                           tlm      = m_model.GetTimeline();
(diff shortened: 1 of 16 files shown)


src/view/src/rocprofvis_data_provider.h
```diff
@@ -224,14 +224,10 @@ class DataProvider
...
-     * indexes of m_track_metadata.
-     * @param track_id: The id of the track to move
-     * @param index: The desired index of the track.
-     * @return: True if operation is successful.
-     */
-    bool SetGraphIndex(uint64_t track_id, uint64_t index);
+    // View-only reorder: sets each TrackInfo::index from its position in
+    // ordered_track_ids (a full permutation) and fires the metadata-changed
+    // callback. Does not touch the controller's graph order.
+    bool SetTrackDisplayOrder(const std::vector<uint64_t>& ordered_track_ids);
```
drchen-amd: I think naming this `SetTrackIndex` would be better. The `TrackInfo` struct that this function modifies has no concept of display order, it only keeps an index.

src/view/src/rocprofvis_project.h
```diff
@@ -153,6 +153,8 @@ constexpr const char* JSON_KEY_TIMELINE_BOOKMARK_Z       = "z";
 
 constexpr const char* JSON_KEY_TIMELINE_TRACK                    = "tracks";
 constexpr const char* JSON_KEY_TIMELINE_TRACK_ORDER              = "order";
+constexpr const char* JSON_KEY_TIMELINE_SORT_MODE               = "sort_mode";
+constexpr const char* JSON_KEY_TIMELINE_CUSTOM_ORDER            = "custom_order";
```
drchen-amd: Do you need this new bespoke `JSON_KEY_TIMELINE_CUSTOM_ORDER` or can you repurpose existing `JSON_KEY_TIMELINE_TRACK_ORDER`? It looks like both are list of track Ids. 

On save; old one dumps the active track order to file, new one dumps `m_custom_order`. What if you change old one to dump `m_custom_order`?

The loading path should just work since both new and old update `m_custom_order`, and it is only applied to the timeline if sort mode is custom.

src/view/src/rocprofvis_timeline_view.h
```diff
@@ -271,6 +311,16 @@ class TimelineView : public RocWidget
...
+
+    // Track sort state. m_default_order is captured once when the trace loads;
+    // m_custom_order is the single remembered manual ordering. Topology order is
+    // fetched on demand from m_topology_order_provider; m_topology_sort_pending is
+    // set when a topology sort is requested before that provider can satisfy it.
+    TrackSortMode                          m_sort_mode;
+    std::vector<uint64_t>                  m_default_order;
+    std::vector<uint64_t>                  m_custom_order;
+    bool                                   m_topology_sort_pending;
+    std::function<std::vector<uint64_t>()> m_topology_order_provider;
```
drchen-amd: I think its odd that this is a function when all it does is get `m_track_order` from `TrackTopology`.

Its fine, but I think this can just be a const ptr to `m_track_order`.

drchen-amd:
I also think some sort of UI element would improve discoverability. Maybe something like
image

src/view/src/rocprofvis_timeline_view.h
```diff
@@ -83,6 +89,15 @@ class TimelineView : public RocWidget
...
 public:
+    // How the timeline track order is derived. Persisted as an int, so keep the
+    // explicit values stable.
+    enum class TrackSortMode
+    {
+        kTopology = 0,  // Match the sidebar topology tree order.
+        kDefault  = 1,  // Natural load order captured when the trace opened.
+        kCustom   = 2,  // The user's manual (drag-reordered) order; one remembered slot.
+    };
+
```
drchen-amd: This is not needed outside of current class so it can be private.

src/view/src/rocprofvis_timeline_view.h
```diff
@@ -128,7 +143,28 @@ class TimelineView : public RocWidget
...
+    // Reorder the tracks according to the given sort mode. If topology order isn't
+    // available yet (the tree is still building at load), the sort is deferred and
+    // applied once SetTopologyOrder() provides it.
+    void          SortTracksBy(TrackSortMode mode);
+    TrackSortMode GetSortMode() const { return m_sort_mode; }
+    bool          HasCustomOrder() const { return !m_custom_order.empty(); }
+    // Points at the sidebar-tree (topology) order owned by TrackTopology. Wired by
+    // the owning TraceView; the pointee stays valid for the view's lifetime.
+    void          SetTopologyOrder(const std::vector<uint64_t>* order);
+
```
drchen-amd, line 157: Only `SetTopologyOrder` has external callers, the rest can be private.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -2457,6 +2633,48 @@ TimelineView::RenderTrackStats(float available_width)
...
+    }
+
+    // Sort menu, shared by the button above and a right-click anywhere on the
+    // header. Pull spacing from the base style; the histogram strip's child has zero
+    // padding, which would otherwise leave the popup cramped.
+    const ImGuiStyle& base = m_settings.GetDefaultStyle();
+    ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, base.WindowPadding);
+    ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, base.ItemSpacing);
+    ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, base.FramePadding);
+    if(ImGui::BeginPopupContextWindow("##track_sort_ctx",
```
drchen-amd: I think now that you have the nice arrow button, the click region to open the menu can be bounded to the button itself rather than the entire stats block.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -2311,6 +2317,176 @@ TimelineView::MakeGraphView()
...
+void
+TimelineView::SetTopologyOrder(const std::vector<uint64_t>* order)
+{
+    m_topology_order = order;
+}
+
+void
+TimelineView::SortTracksBy(TrackSortMode mode)
+{
+    m_sort_mode = mode;
```
drchen-amd: I don't think it is good to write `m_sort_mode` first thing because it could fail downstream for various reasons especially in `ApplyTrackOrder`. 

It would be better to write after everything is confirmed okay.

src/view/src/rocprofvis_timeline_view.h
```diff
@@ -128,7 +134,38 @@ class TimelineView : public RocWidget
...
+        kTopology = 0,  // Match the sidebar topology tree order.
+        kDefault  = 1,  // Natural load order captured when the trace opened.
+        kCustom   = 2,  // The user's manual (drag-reordered) order; one remembered slot.
+    };
+
+    // Reorder the tracks according to the given sort mode. If topology order isn't
+    // available yet (the tree is still building at load), the sort is deferred and
+    // applied once SetTopologyOrder() provides it.
+    void          SortTracksBy(TrackSortMode mode);
+    TrackSortMode GetSortMode() const { return m_sort_mode; }
```
drchen-amd: This is unused.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -286,47 +286,41 @@ DataProvider::SetRequestProgressUpdateCallback(
...
-                        tlm.GetMutableTrackMetadata()[id].index = i;
-                    }
-                }
-            }
-            if(m_track_metadata_changed_callback)
-            {
-                m_track_metadata_changed_callback(m_model.GetTraceFilePath());
-            }
+            spdlog::warn("SetTrackIndex ignored: unknown track id {}", track_id);
+            return false;
```
drchen-amd, line 318: If anything returns here, you will end up with a part of the list with updated indexes, and part of the list with old indexes. 

I would confirm everything is valid before touching any indexes.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -2311,6 +2318,180 @@ TimelineView::MakeGraphView()
...
+        {
+            const TrackInfo* metadata = tlm.GetTrack(track->GetID());
+            if(metadata && metadata->index < reordered.size())
+            {
+                reordered[metadata->index] = track;
+            }
+        }
+    }
+    *m_tracks = std::move(reordered);
+}
```
drchen-amd, line 2459: This code was originally embedded in `Update` to defer it from happening inside render loop. I think moving it to a function is risky because this code modifies the track list, and there are many places that traverse this list in their render loops. Today this is not causing a problem, but going forward callers of this function must take care to not do so while a reader of track list is mid iteration.

It would to preferable to keep this in `Update` and have callers trigger in a deferred manner via setting a flag instead of directly calling.

tomk-amd: I think this is ok in a seprate function, it makes the code more readable.  The function is already private so only TimelineView can do something dangerous.  I would just put a comment saying that this must only be called from the update() function.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -2335,6 +2342,182 @@ TimelineView::MakeGraphView()
...
+    }
+
+    RebuildTrackVectorFromMetadata();
+    // Force the position map / layout to recompute on the next Update().
+    m_resize_activity = true;
+    return true;
+}
+
+// Replaces m_tracks; only call from Update(), never mid-render, since the render
+// loops iterate this list.
```
drchen-amd, line 2441: This comment has already been violated because the current flow is:
container->Render()
RenderHistogram()
RenderTrackStats()
RenderTrackSortMenu()
SortTracksBy()
ApplyTrackOrder()
RebuildTrackVectorFromMetadata()

drchen-amd (dismissed):

src/view/src/rocprofvis_timeline_view.cpp
tomk-amd: Instead of `SortTracksBy()` calling `ApplyTrackOrder()` and violating the "rule", it instead could set an extra member variable like m_pending_sort_mode.   Then in the `Update()` function, if `m_pending_sort_mode != m_sort_mode` do `ApplyTrackOrder()`.

drchen-amd: While we are here, it would also be nice if this flow were restructured such that the hierarchy makes sense because track stats and sorting do not belong inside "histogram". `RenderHistogram` could be renamed to something like `RenderHeader`, then from this further split into:
- `RenderHeader`
  - `RenderHistogram`
  - `RenderTrackInfo`
    - `RenderTrackStats`
    - `RenderTrackSortMenu`

drchen-amd:
We offer option to disable the histogram, and since this feature is inside histogram, it will be inaccessible if the histogram is disabled. I think we should at least make the sorting menu accessible in the track meta-area context menu if the histogram is hidden. The show histogram option should also be renamed if it is going to affect more than just the histogram.

drchen-amd (approved):

PR #983, opened 2026-08-06
View: Make the measurement duration label draggable

dhingora-amd:
## Motivation
The measurement tool's duration label was pinned in place, could be clipped off
short traces, and leaked input and visuals through to the layers below. This
makes it a draggable, always-on-screen readout and fixes those issues.

## Technical Details
In `TimelineView` (`src/view/src/rocprofvis_timeline_view.cpp` / `.h`):

- **Draggable label.** The duration label can be dragged vertically and the
  connecting line follows it. New session-only state `m_measure_label_offset_y` /
  `m_dragging_measure_label` and `HandleMeasurementLabelInput()` handle the drag;
  the offset is clamped to the visible area and reset when the measurement clears.
- **Input fix.** Measurement is drawn in a viewport-sized `NoInputs` overlay,
  with drag input in a dedicated child sized to the label. Lower layers are
  blocked only while grabbing (not on hover), and input runs before draw so a
  drag lands the same frame.
- **Clipping fix.** The overlay pushes a clip rect over the full visible graph
  area (instead of clipping at `m_track_height_sum`), so the label stays visible
  on short traces; ruler lines run to `max(last track, viewport bottom)`.
- **Layering.** Labels are drawn above the annotation anchors and on top of all
  lines, with an opaque fill so lines don't bleed through. Left ruler timestamp
  sits at the top, right at the bottom, to avoid overlap when the rulers are close.
- **Cleanup.** Padding/rounding promoted to shared `LABEL_PADDING` /
  `LABEL_ROUNDING` constants.


tomk-amd:
After playing with this feature here are some general comments / observations:
- The measure mode and range mode labels should interact the same way.  (It was a wrong decision to make the range selection label freely dragable.. my bad 👎 )
- I think the range selection label should only be draggable in the vertical direction just as the measure mode one.
- But ... I'm still not convinced that this label should even be present on the time line..

Ideally we could find a spot to show: Selection Start: 33333 ns  Selection End: 44444  Selection Span: 11111 

I think we could fit it in the histogram area like this?
image

and it would work well with the changes in https://github.com/ROCm/roc-optiq/pull/995

Thoughts?

dhingora-amd:
> After playing with this feature here are some general comments / observations:
> 
> * The measure mode and range mode labels should interact the same way.  (It was a wrong decision to make the range selection label freely dragable.. my bad 👎 )
> * I think the range selection label should only be draggable in the vertical direction just as the measure mode one.
> * But ... I'm still not convinced that this label should even be present on the time line..
> 
> Ideally we could find a spot to show: Selection Start: 33333 ns Selection End: 44444 Selection Span: 11111
> 
> I think we could fit it in the histogram area like this? image
> 
> and it would work well with the changes in #995
> 
> Thoughts?

This would be an ideal solution but I am worried on how it would look if the range is very short. Since the histogram does not adapt to view, if the range is really short, the span label would have to go above, which would look off, and the left and right span would have to be further from the lines... I can attempt it and see how it would look, but I am having a bit of struggle imagining how to format it in the case of a short range which requires all 3 numbers to be extremely close to each other.

tomk-amd:
> 
> This would be an ideal solution but I am worried on how it would look if the range is very short. Since the histogram does not adapt to view, if the range is really short, the span label would have to go above, which would look off, and the left and right span would have to be further from the lines... I can attempt it and see how it would look, but I am having a bit of struggle imagining how to format it in the case of a short range which requires all 3 numbers to be extremely close to each other.

Yes I realize that will be inevitable.. so in that case the delta label could either be topmost and overlap (hide) the other two.. or get pushed down and overlap histogram?  Even the start and end labels will run into problems too if the selection is narrow and near the edge of the view.

dhingora-amd:
> Yes I realize that will be inevitable.. so in that case the delta label could either be topmost and overlap (hide) the other two.. or get pushed down and overlap histogram? Even the start and end labels will run into problems too if the selection is narrow and near the edge of the view.

In the case of overlapping/hiding, what should we do if the user wants to see the start and end? Covering the histogram seems like it would be defeating the point of the histogram....

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -659,20 +682,46 @@ TimelineView::RenderMeasurement(ImDrawList* draw_list, ImVec2 window_position)
...
+        }
+        if(hovered && ImGui::IsMouseClicked(ImGuiMouseButton_Left))
+        {
+            m_dragging_measure_label = true;
+        }
+        if(m_dragging_measure_label && ImGui::IsMouseDown(ImGuiMouseButton_Left))
+        {
+            m_measure_label_offset_y += ImGui::GetIO().MouseDelta.y;
+        }
+        if(!ImGui::IsMouseDown(ImGuiMouseButton_Left))
```
tomk-amd: This will need "z-level / focus" detection, see video in main conversation

tomk-amd:
https://github.com/user-attachments/assets/b44287ab-7665-4edd-9b4d-a86548f3344c

Z-level mouse focus needs to be added

tomk-amd:
Things go horribly wrong if tracks are shown / hidden while measurement is enabled

image

The view extends (vertical scrollbar present) beyond the track height.

Might have something to do with this:

InvisibleButton can grow Graph View Main content height

The button is placed in last frame’s screen rect, which on short traces is often below the last track (that is the clip fix). Restoring the cursor does not restore CursorMaxPos, so ImGui can treat empty space as content.

m_content_max_y_scroll is sampled before the button, so the next frame’s GetScrollMaxY() can become non-zero on a trace that previously did not scroll. Dragging the label to the bottom of a short trace is the likely trigger (phantom scroll, view jump).

Put the hit target in a non-scrolled, viewport-sized overlay (or save/restore CursorMaxPos if you already include imgui_internal.h). Please verify: short trace, complete a measurement, drag the duration label to the bottom, see if a vertical scrollbar appears.

tomk-amd:
Also this is a pre-existing problem (not this branch / PR) but the measurement "layer" should have a higher z-index than the annotation anchor:

image

It is ok for the annotation note (expanded window) to float at the top level

image


but the anchor should be above events and sample lines, but below pretty much every thing else...

tomk-amd:
Finally another comment from AI review but it is real, hovering the measurement label shouldn't block scrollwheel (zoom) or WASD (or other keyboard interaction) -- see below:


While the cursor sits on the duration label (it defaults to viewport center), wheel zoom, horizontal wheel pan, keyboard pan/zoom, ESC, and freehand ruler drags all stop. Annotations already do something similar, but this label is a small box in the middle of the graph, so it is easy to rest on.

Keep the cursor and RequestLayerFocus on hover; set m_stop_user_interaction only when the item is active (or on left mouse down). That still blocks scrubber/selection leak-through, which uses hand-rolled IsMouseClicked, without eating wheel/hotkeys on hover.

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -527,6 +541,43 @@ TimelineView::AutoScrollForAnnotationDrag(ImVec2 content_origin)
...
+                      ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoScrollWithMouse);
+    ImGui::InvisibleButton("##measure_duration_label", size);
+    const bool hovered = ImGui::IsItemHovered();
+    const bool active  = ImGui::IsItemActive();
+    ImGui::EndChild();
+    ImGui::PopStyleVar();
+
+    m_dragging_measure_label = active;
+    if(hovered || active)
+    {
```
tomk-amd: m_dragging_measure_label is set / reset but never actually checked against, is it needed?

tomk-amd (commented):
Please fix the PR desc as well, as this PR has evolved since it was created.  Scope is now to just fix the measure tool behaviour.

tomk-amd (approved):

PR #985, opened 2026-08-07
View: Add time format preview and format hints to settings

dhingora-amd:
## Motivation

The time format setting (Settings → Time) gave no indication of what each
option looked like, and the "Timecode" options were ambiguous. This makes
the setting self-explanatory.

## Technical Details

- Added a live preview under the dropdown that formats a sample duration
  in the selected `TimeFormat`, styled to match the font preview above.
- Reuses the existing `nanosecond_to_formatted_str(...)` helper.
- Annotated the timecode options with their layout, e.g.
  `Timecode (hh:mm:ss.ns)`.

### DIFF
--- a/src/view/src/rocprofvis_settings_panel.cpp
+++ b/src/view/src/rocprofvis_settings_panel.cpp
@@ -21,8 +21,10 @@ constexpr float kHotkeyCategoryColWidth   = 90.0f;
 constexpr float kHotkeyBindingColWidth    = 120.0f;
 constexpr float kLogViewerInputWidth      = 160.0f;
 
-// Sample duration (1h 23m 45.123456789s) shown in the time-format preview.
-constexpr double kPreviewSampleNs = 5025123456789.0;
+// Sample duration shown in the time-format preview: 12m 34.123456789s. Uses zero
+// hours so Condensed Timecode (which trims leading zero hour/minute groups)
+// renders differently from Timecode.
+constexpr double kPreviewSampleNs = 754123456789.0;
 
 namespace RocProfVis
 {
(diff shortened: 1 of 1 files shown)


src/view/src/rocprofvis_settings_panel.cpp
```diff
@@ -308,6 +312,30 @@ SettingsPanel::RenderUnitOptions()
         m_settings_changed = true;
     }
     PopComboStyles();
+
+    // Time format preview, styled to match the font preview above.
+    const TimeFormat selected_format = m_usersettings.unit_settings.time_format;
+    const bool       is_timecode     = selected_format == TimeFormat::kTimecode ||
+                                 selected_format == TimeFormat::kTimecodeCondensed;
+    const std::string preview =
+        nanosecond_to_formatted_str(kPreviewSampleNs, selected_format, !is_timecode);
```
drchen-amd: This can just be `nanosecond_to_formatted_str(kPreviewSampleNs, selected_format, true)`. The timecodes ignore last parameter anyway.

src/view/src/rocprofvis_settings_panel.cpp
```diff
@@ -289,15 +293,15 @@ SettingsPanel::RenderUnitOptions()
...
                             2 * style.FramePadding.x +
                             ImGui::GetFrameHeightWithSpacing());
     int time_format_index = static_cast<int>(m_usersettings.unit_settings.time_format);
     // Options must match TimeFormat enum
     PushComboStyles();
     if(ImGui::Combo("##time_format", &time_format_index,
-                    "Timecode\0"
-                    "Condensed Timecode\0"
+                    "Timecode (hh:mm:ss.ns)\0"
+                    "Condensed Timecode (mm:ss.ns)\0"
```
drchen-amd: I think condensed timecode should also list `(hh:mm:ss.ns)` because it will show hours and minutes provided they are non-zero. Unless you know of a clean way to indicate something may be there.

drchen-amd (approved):

PR #987, opened 2026-08-07
View: Add "Z" hotkey to zoom to the current selection

dhingora-amd:
## Motivation

Add a keyboard shortcut to zoom the timeline to the current selection, so users
can do it without reaching for the mouse.

## Technical Details

- New rebindable **"Zoom to Selection"** hotkey (default: `Z`) in `HotkeyManager`;
  auto-listed and persisted via Settings > Hotkeys.
- `TraceView::HandleHotKeys()` fits the view to the region/time-range selection
  (falls back to selected events), preserving vertical scroll; warns if nothing
  is selected.

### DIFF
--- a/src/view/src/rocprofvis_trace_view.cpp
+++ b/src/view/src/rocprofvis_trace_view.cpp
@@ -466,33 +466,9 @@ TraceView::HandleHotKeys()
 
     if(hk.WasActionTriggered(HotkeyActionId::kZoomToSelection))
     {
-        if(m_timeline_view && m_timeline_selection)
+        if(m_timeline_view)
         {
-            double start_ns  = 0.0;
-            double end_ns    = 0.0;
-            bool   has_range = false;
-
-            if(m_timeline_selection->HasValidTimeRangeSelection())
-            {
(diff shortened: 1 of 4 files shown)


src/view/src/rocprofvis_trace_view.cpp
```diff
@@ -463,6 +463,38 @@ TraceView::HandleHotKeys()
...
+            if(has_range && end_ns > start_ns)
+            {
+                m_timeline_view->MoveToPosition(
+                    start_ns, end_ns, m_timeline_view->GetViewCoords().y, false);
+            }
+            else
+            {
+                NotificationManager::GetInstance().Show(
+                    "No selection to zoom to.", NotificationLevel::Warning);
+            }
```
tomk-amd: Instead of re-implementing, call ZoomToTimeRangeSelection() instead.. I don't think the notification is needed, silent failure is fine in this case.

tomk-amd (changes requested):

tomk-amd (approved):

PR #991, opened 2026-08-11
Search Enhancements

drchen-amd:
Include memory events in search for supported file formats, adds an "Advanced" search mode that allows for more user control.

### DM
- Implement search query for memory allocate events for rocprof files.
  - Uses modified method, because unlike other types...
     - Their name strings are not stored in the string table.
     - Their base operation query retrieves the name as string rather than id.
  - Approach - Match each search term against name permutations (hard coded in `Builder::mem_alloc_types`), then generate WHERE clause for confirmed matches (ex `name IN ("ALLOC", "FREE")`)
 - Implement search query for memory copy for rocprof files.
 - Implement new search parameters for all file types :
    - enable/disable searching of event category in addition to event name.
    - enable/disable inclusion of events that match a subset of a multi-term search

### Controller
- Expose new search parameters `kRPVControllerTableArgsStringTableFiltersIncludeCategory`, `kRPVControllerTableArgsStringTableFiltersPartialMatching`.

### View
- Add UI for configuring search parameters.
- InfiniteScrollTable
  - Reduced column indexing to those strictly required.
  - Remove usage of `ImGuiTableColumnFlags_DefaultHide` to hide columns. `ImGuiTableColumnFlags_Disabled` is enough.
    - Flags prefixed with `Default` are only read once by Imgui and cached thereafter, even if caller sets them to a different set of columns, only the original initial set is honored. This is why columns were observed to set themselves to hidden when the set of columns changed, and why pushing a different ID could workaround it by forcing the initial set to be read again.
- Fix pre-made range selection being interactable through windows/popups. 
- Generalized `AccentButton` gui helper into `ColoredButton`

### DIFF
--- a/src/controller/src/system/rocprofvis_controller_table_system_search.cpp
+++ b/src/controller/src/system/rocprofvis_controller_table_system_search.cpp
@@ -12,6 +12,8 @@ namespace Controller
 EventSearchTable::EventSearchTable(uint64_t id)
 : SystemTable(id)
 , m_include_substrings(true)
+, m_include_category(false)
+, m_partial_matching(false)
 {
 }
 
@@ -24,6 +26,8 @@ void EventSearchTable::Reset()
     SystemTable::Reset();
     m_string_table_filters.clear();
     m_include_substrings = true;
+    m_include_category = false;
+    m_partial_matching = false;
 }
 
 rocprofvis_result_t
@@ -39,6 +43,8 @@ EventSearchTable::UnpackArguments(Arguments& args, TableArguments*& out) const
     {
         uint64_t num_string_table_filters = 0;
         uint64_t include_substrings = 1;
+        uint64_t include_category = 0;
+        uint64_t partial_matching = 0;
         std::vector<std::string> string_table_filters;
         uint64_t table_type = static_cast<uint64_t>(kRPVControllerTableTypeSearchResults);
 
@@ -70,11 +76,21 @@ EventSearchTable::UnpackArguments(Arguments& args, TableArguments*& out) const
             {
                 result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersIncludeSubstrings, 0, &include_substrings);
             }
+            if(result == kRocProfVisResultSuccess)
+            {
+                result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersIncludeCategory, 0, &include_category);
+            }
+            if(result == kRocProfVisResultSuccess)
+            {
+                result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersPartialMatching, 0, &partial_matching);
+            }
         }
         if(result == kRocProfVisResultSuccess)
         {
             search_out->m_string_table_filters = std::move(string_table_filters);
             search_out->m_include_substrings = include_substrings != 0;
+            search_out->m_include_category = include_category != 0;
+            search_out->m_partial_matching = partial_matching != 0;
         }
     }
 	return result;
@@ -91,6 +107,8 @@ EventSearchTable::GetCurrentArguments(TableArguments*& out) const
     SystemTable::GetCurrentArguments(out);    
     search_out->m_string_table_filters = m_string_table_filters;
     search_out->m_include_substrings = m_include_substrings;
+    search_out->m_include_category = m_include_category;
+    search_out->m_partial_matching = m_partial_matching;
 }
 
 void
@@ -100,13 +118,15 @@ EventSearchTable::SetCurrentArguments(TableArguments& in)
     SystemTable::SetCurrentArguments(search_in);
     m_string_table_filters = search_in.m_string_table_filters;
     m_include_substrings = search_in.m_include_substrings;
+    m_include_category = search_in.m_include_category;
+    m_partial_matching = search_in.m_partial_matching;
 }
 
 bool
 EventSearchTable::ArgumentsChanged(SystemTableArguments& in) const
 {
     EventSearchTableArguments& search_in = (EventSearchTableArguments&)in;
-    return SystemTable::ArgumentsChanged(search_in) || m_string_table_filters != search_in.m_string_table_filters || m_include_substrings != search_in.m_include_substrings;
+    return SystemTable::ArgumentsChanged(search_in) || m_string_table_filters != search_in.m_string_table_filters || m_include_substrings != search_in.m_include_substrings || m_include_category != search_in.m_include_category || m_partial_matching != search_in.m_partial_matching;
 }
 
 rocprofvis_result_t
@@ -133,7 +153,7 @@ EventSearchTable::BuildQuery(rocprofvis_dm_database_t db, TableArguments& args, 
                                                     static_cast<rocprofvis_db_num_of_tracks_t>(arguments.m_tracks.size()), arguments.m_tracks.data(),
                                                     arguments.m_where.c_str(),
                                                     static_cast<rocprofvis_dm_num_string_table_filters_t>(string_table_filters_ptr.size()), string_table_filters_ptr.data(),
-                                                    arguments.m_include_substrings,
+                                                    arguments.m_include_substrings, arguments.m_include_category, arguments.m_partial_matching,
                                                     sort_column, (rocprofvis_dm_sort_order_t)arguments.m_sort_order,
                                                     count_only ? 0 : count, count_only ? 0 : index, count_only, out);
     return result;
--- a/src/model/src/database/rocprofvis_db_query_manager.cpp
+++ b/src/model/src/database/rocprofvis_db_query_manager.cpp
@@ -525,7 +525,9 @@ QueryManager::BuildEventSearchQuery(
         rocprofvis_dm_index_t op = TABLE_QUERY_UNPACK_OP_TYPE(ops[i]);
         if (num_string_table_filters > 0)
         {
-            if (string_filter_result == kRocProfVisDmResultSuccess && string_id_filter_map.count((rocprofvis_dm_event_operation_t)op) > 0 && !GetEventOperationQuery((rocprofvis_dm_event_operation_t)op).empty())
+            if (string_filter_result == kRocProfVisDmResultSuccess && 
+                string_id_filter_map.count((rocprofvis_dm_event_operation_t)op) > 0 && 
+                !GetEventOperationQuery((rocprofvis_dm_event_operation_t)op).empty())
             {
                 const auto& filters = string_id_filter_map.at((rocprofvis_dm_event_operation_t)op);
                 for (auto it = filters.begin(); it != filters.end(); ++it)
--- a/src/model/src/database/rocprofvis_db_query_manager.cpp
+++ b/src/model/src/database/rocprofvis_db_query_manager.cpp
@@ -511,22 +511,21 @@ QueryManager::BuildEventSearchQuery(
     rocprofvis_dm_timestamp_t start, rocprofvis_dm_timestamp_t end,
     rocprofvis_db_num_of_tracks_t num, rocprofvis_db_track_selection_t ops,
     rocprofvis_dm_charptr_t where,
-    rocprofvis_dm_num_string_table_filters_t num_string_table_filters,
-    rocprofvis_dm_string_table_filters_t     string_table_filters,
-    bool include_substring,
+    rocprofvis_dm_num_string_table_filters_t num_string_table_filters, rocprofvis_dm_string_table_filters_t string_table_filters,
+    bool include_substring, bool include_category, bool partial_matching,
     rocprofvis_dm_charptr_t sort_column, rocprofvis_dm_sort_order_t sort_order,
     uint64_t max_count, uint64_t offset, bool count_only, rocprofvis_dm_string_t& query)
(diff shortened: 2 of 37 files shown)


src/view/src/widgets/rocprofvis_gui_helpers.cpp
```diff
@@ -645,14 +645,27 @@ AccentButton(const char* label, ImVec2 size, SettingsManager* settings)
     {
         settings = &SettingsManager::GetInstance();
     }
-    ImGui::PushStyleColor(ImGuiCol_Button, settings->GetColor(Colors::kAccent));
-    ImGui::PushStyleColor(ImGuiCol_ButtonHovered,
-                          settings->GetColor(Colors::kAccentHover));
-    ImGui::PushStyleColor(ImGuiCol_ButtonActive,
-                          settings->GetColor(Colors::kAccentActive));
-    ImGui::PushStyleColor(ImGuiCol_Text, settings->GetColor(Colors::kTextOnAccent));
+    return ColoredButton(label, settings->GetColor(Colors::kAccent),
```
amokiche-amd, line 648: ColorredButon don't use the size parameter, if it intentional it not obvious why

src/controller/src/system/rocprofvis_controller_table_system_search.cpp
```diff
@@ -70,11 +76,21 @@ EventSearchTable::UnpackArguments(Arguments& args, TableArguments*& out) const
...
                 result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersIncludeSubstrings, 0, &include_substrings);
             }
+            if(result == kRocProfVisResultSuccess)
+            {
+                result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersIncludeCategory, 0, &include_category);
+            }
+            if(result == kRocProfVisResultSuccess)
+            {
+                result = args.GetUInt64(kRPVControllerTableArgsStringTableFiltersPartialMatching, 0, &partial_matching);
+            }
```
amokiche-amd, line 86: Should these arguments be mandatory? If so, and this is intentional, then there are no questions.

src/model/src/database/rocprofvis_db_trace_processor.h
```diff
@@ -129,6 +129,8 @@ class GoogleTraceProcessor : public QueryManager
             rocprofvis_dm_num_string_table_filters_t num_string_table_filters, 
             rocprofvis_dm_string_table_filters_t string_table_filters,
             bool include_substring,
+            bool include_category,
+            bool partial_matching,
```
amokiche-amd, line 133: Name of the variable litle bit confused, becaouse it not obvious what the difference between include_substring and partitial matshing?

src/model/src/database/rocprofvis_db_query_manager.cpp
```diff
@@ -511,22 +511,21 @@ QueryManager::BuildEventSearchQuery(
...
-    rocprofvis_dm_result_t string_filter_result = BuildTableStringIdFilter(num_string_table_filters, string_table_filters, include_substring, string_id_filter_map);
+    rocprofvis_dm_result_t string_filter_result = BuildTableStringIdFilter(num_string_table_filters, string_table_filters, include_substring, include_category, partial_matching, string_id_filter_map);
     slice_query_map_array.resize(num);
     for(int i = 0; i < num; i++)
     {
         rocprofvis_dm_index_t op = TABLE_QUERY_UNPACK_OP_TYPE(ops[i]);
         if (num_string_table_filters > 0)
         {
-            if (string_filter_result == kRocProfVisDmResultSuccess && string_id_filter_map.count((rocprofvis_dm_event_operation_t)op) > 0)
+            if (string_filter_result == kRocProfVisDmResultSuccess && string_id_filter_map.count((rocprofvis_dm_event_operation_t)op) > 0 && !GetEventOperationQuery((rocprofvis_dm_event_operation_t)op).empty())
```
amokiche-amd: I still think that c++ cast more clear. And we have a plan to refactor our backend and it will be better to start use it where it possible
Also you can cast it just once and use the result insted cast it twice
and it too large line will be better to make it smaller and move par to another line

src/model/src/database/rocprofvis_db_query_manager.cpp
```diff
@@ -511,22 +511,21 @@ QueryManager::BuildEventSearchQuery(
...
             {
                 auto filters = string_id_filter_map.at((rocprofvis_dm_event_operation_t)op);
```
amokiche-amd:
```suggestion
                const auto& filters = string_id_filter_map.at((rocprofvis_dm_event_operation_t)op);
```

Not your changes but will be better fix it. Here we copy whole intire map std::unordered_map<uint32_t, std::string>

src/model/src/database/rocprofvis_db_rocprof.cpp
amokiche-amd: also unnecesary copy
```suggestion
            for (const auto& string_id_obj : string_id_array)
```

src/model/src/database/rocprofvis_db_rocprof.cpp
```diff
@@ -1755,51 +1756,102 @@ rocprofvis_dm_result_t RocprofDatabase::BuildTableStringIdFilter( rocprofvis_dm_
...
+            {
+                mem_alloc_types += mem_alloc_types.empty() ? '\'' + type + '\'' : ", '" + type + '\'';
+            }
+        }
 
-
-        if (string_ids.size() > 0)
+        if(mem_alloc_types.size())
+        {                
+            for(GuidInfo& instance : DbInstances())
```
amokiche-amd, line 1817:
```suggestion
            for(const GuidInfo& instance : DbInstances())
```

src/view/src/rocprofvis_event_search.h
```diff
@@ -16,36 +15,52 @@ class EventSearch : public InfiniteScrollTable
 {
 public:
     EventSearch(DataProvider& dp, std::shared_ptr<TimelineSelection> timeline_selection);
+    ~EventSearch();
```
amokiche-amd:
```suggestion
    ~EventSearch() override;
```

src/model/src/datamodel/rocprofvis_dm_trace.cpp
```diff
@@ -802,7 +803,7 @@ rocprofvis_dm_result_t Trace::SearchStringIndices(rocprofvis_dm_num_string_table
                                 return std::tolower(c1) == std::tolower(c2);
                             });
             }
-            if(!match)
+            if(match == partial_matching)
             {
                 break;
             }
```
amokiche-amd, line 809: Now wi have different behavor for search in the dm_trace.cpp and here. It looks like the expected behavor from here is the right but it looks like potential bug in dn_trace.
Here the example that prodused different search result with the same options
<html>
<body>
<h2>Example producing different results</h2><p>Consider this event:</p><pre dir="ltr"><code>name:     hipLaunchKernel
category: Runtime API</code></pre><p>Search terms:</p><pre dir="ltr"><code>kernel
runtime</code></pre><p>UI input:</p><pre dir="ltr"><code>"kernel""runtime"</code></pre><p>Options:</p><pre dir="ltr"><code>Contains
AND
Search Event Categories = enabled</code></pre><h3>Perfetto result</h3><p>Generated predicate:</p><pre dir="ltr"><code>(name LIKE '%kernel%' OR category LIKE '%kernel%')
AND
(name LIKE '%runtime%' OR category LIKE '%runtime%')</code></pre><p>Evaluation:</p><pre dir="ltr"><code>First term, "kernel":
    name contains "kernel"       = true
    category contains "kernel"   = false
    true OR false                = true

Second term, "runtime":
    name contains "runtime"      = false
    category contains "runtime"  = true
    false OR true                = true

Final:
    true AND true = true</code></pre><p>Perfetto returns the event.</p><h3><code dir="ltr">dm_trace</code> result</h3><p>The trace string table contains separate entries:</p><pre dir="ltr"><code>sv = "hipLaunchKernel"
sv = "Runtime API"</code></pre><p>For <code dir="ltr">"hipLaunchKernel"</code>:</p><pre dir="ltr"><code>contains "kernel"  = true
contains "runtime" = false

true AND false = false</code></pre><p>For <code dir="ltr">"Runtime API"</code>:</p><pre dir="ltr"><code>contains "kernel" = false</code></pre><p>The loop stops immediately because AND mode encountered a false comparison:</p><pre dir="ltr"><code>false AND ... = false</code></pre><p>Neither individual string matches both terms, so neither string index is returned.</p><p>Rocprof/RocPD then have no matching name/category IDs for this event.</p><p>Final outcome:</p><div><div><div>
Adapter | Result
-- | --
Perfetto | Event is returned
Rocprof/RocPD through dm_trace | Event is not returned

</div></div></div><p>The difference can be summarized as:</p><pre dir="ltr"><code>Perfetto:
    all terms must match somewhere in the event's enabled fields

dm_trace:
    one individual string must match all terms</code></pre>
</body>
</html>

src/model/src/database/rocprofvis_db_rocprof.cpp
```diff
@@ -1755,51 +1756,102 @@ rocprofvis_dm_result_t RocprofDatabase::BuildTableStringIdFilter( rocprofvis_dm_
...
+                if(match == partial_matching)
+                {
+                    break;
+                }
+            }
+            if(match)
+            {
+                mem_alloc_types += mem_alloc_types.empty() ? '\'' + type + '\'' : ", '" + type + '\'';
+            }
+        }
```
amokiche-amd, line 1813: This code looks like line by line copy of the rocprofvis_dm_trace.cpp:784 I know this is the diferent modules but may be wee need any comon helpers library? Becaouse it looks like potential problem in this code that described bellow and we have to copies of the same code

src/view/src/rocprofvis_event_search.cpp
```diff
@@ -28,22 +34,56 @@ EventSearch::EventSearch(DataProvider&                      dp,
...
+
+EventSearch::~EventSearch()
+{
+    EventManager::GetInstance()->Unsubscribe(
+        static_cast<int>(RocEvents::kTimelineTimeRangeChanged),
+        m_time_range_changed_token);
+}
 
 void
 EventSearch::Update()
```
amokiche-amd, line 73: It seems to me that the function turned out to be too large and it would be better to break it into smaller functions for the sake of cleanliness and accuracy of the code.

amokiche-amd (commented):
Overall, everything looks great, but there is one potential logical problem, a few suggestions for improving code readability, and some minor nitpicks.

src/model/src/datamodel/rocprofvis_dm_trace.cpp
tomk-amd: While compariing against Perfetto path is interesting, our Perfetto support right now is experimental. (To the point that it might be removed -- or at least implemented by Profiler Hub and not our data model).   So these differences are out-of-scope for this PR.

src/model/src/database/rocprofvis_db_rocprof.cpp
drchen-amd: The function I need is not const, hence this was not const. That function could be made const, but I didn't think it was enough of a reason to touch that class for something so minor.

src/controller/src/system/rocprofvis_controller_table_system_search.cpp
drchen-amd: Yes I would like to these to be mandatory.

src/model/src/database/rocprofvis_db_rocprof.cpp
```diff
@@ -1735,71 +1735,123 @@ rocprofvis_dm_result_t RocprofDatabase::SaveTrimmedData(rocprofvis_dm_timestamp_
...
                 {
                     filter[kRocProfVisDmOperationDispatch][it->first] += ") OR K.kernel_id IN (" + k_it->second;
                 }
-                filter[kRocProfVisDmOperationDispatch][it->first] += ") ";
+
+                if(include_category)
+                {
+                    filter[kRocProfVisDmOperationLaunch][it->first] += ") OR E.category_id IN (" + it->second;
+                    filter[kRocProfVisDmOperationDispatch][it->first] += ") OR E.category_id IN (" + it->second;
+                    filter[kRocProfVisDmOperationMemoryAllocate][it->first] = filter[kRocProfVisDmOperationMemoryAllocate][it->first].empty() ? " E.category_id IN (" + it->second: "(" + filter[kRocProfVisDmOperationMemoryAllocate][it->first] + ") OR E.category_id IN (" + it->second + ")";
```
vstempen: Can you format this line? Split it in 3 parts.

src/model/src/database/rocprofvis_db_query_manager.cpp
drchen-amd: I've split this line into multiple lines.

src/model/src/database/rocprofvis_db_trace_processor.h
drchen-amd: I'm open to better names. Originally `partial_matching` was `use_or` and `use_and`, but I didn't want two separates parameters.

tomk-amd (approved):

PR #993, opened 2026-08-11
Harden profiling backend

tomk-amd:
## Motivation

Tightens what a profiler launch can put on the command line.  
This is the first phase of necessary preparation work for supporting other profiling tools like rocprofv3 and Compute.
Ex:  working directory support, needed for running Compute profiler down the line. 

## Technical Details

**Compose the command as an explicit argv.** Backends emit the complete
argument list from `FlattenToExecution` and the controller passes each entry
to the process verbatim. Previously arguments were joined into a string and
re-split on whitespace, so any path or argument containing a space was torn
into several, and the command preview re-derived its own tail and could show
something other than what ran. `target_executable` / `output_directory`
become metadata that never reach the command line; `set_target_args` /
`set_profiler_args` are dropped from the C API. Adds quote-aware
`SplitArguments` and replaces seven positional launch strings with
`ProfilerLaunchSpec`.

**Name tools by enum, not by path.** A launch carries a
`rocprofvis_profiler_tool_t`; the controller maps it to a binary name and
resolves a validated absolute path before exec, so no caller-supplied string
becomes `argv[0]`. A missing tool reports `kRocProfVisResultToolNotFound`
rather than surfacing as exit code 127. `ProfilerSettings::profiler_path` is
deleted -- a ROCm install in a non-standard location uses
`LaunchConfig::tool_directory`, a directory whose filename still comes from
the tool table. Remote launches resolve on the remote host, never locally.

Also in scope: working-directory support (applied to the child only, failing
the run rather than writing output somewhere unasked-for), a fix for the
output getter dropping its last character, and trace-path scraping that
prefers the last labelled `.db` path -- profiling Optiq with Optiq
interleaves the child's own log lines with the tool's.

## Compatibility

- `profiler_path` in an existing settings file is now read by nothing.
- Launch profiles persist `tool` as the enum's integer, so the enum is
  append-only. A profile written with the old `tool_id` string loads with no
  tool selected, and the launcher substitutes the backend's first tool.

## Test plan

- [x] New `roc-optiq-controller-profiler-tests` target (gated on
      `ROCPROFVIS_ENABLE_PROFILER`): 17 cases, 111 assertions, covering argv
      composition, remote shell quoting, env-name rejection, resolution order,
      tool-directory strictness, working-directory behavior, and an
      unresolvable tool failing the launch without spawning. Passing on Linux.
- [x] Windows build and test run.
- [x] Manual: local launch, remote/SSH launch, and a target path containing a
      space, confirming the preview matches the logged command line.

### DIFF
--- a/src/view/src/profiler/rocprofvis_profiler_launcher_dialog.cpp
+++ b/src/view/src/profiler/rocprofvis_profiler_launcher_dialog.cpp
@@ -1237,15 +1237,15 @@ void ProfilerLauncherDialog::OnLaunchClicked()
     // to the orchestrator. The backend scraper resolves the produced trace path
     // from the profiler's stdout (local and remote).
     ProfilerLaunchOrchestrator::LaunchRequest request;
-    request.profiler_type     = ResolveProfilerType();
-    request.profiler_path     = m_execution_cache.profiler_path;
-    request.target_executable = m_config.target.executable;
-    request.target_args       = m_config.target.arguments;
-    request.output_directory  = m_config.target.output_directory;
-    request.profiler_args     = m_execution_cache.profiler_args;
-    request.env_vars          = m_execution_cache.env_vars;
-    request.is_remote         = is_remote;
-    request.auto_load_trace   = m_config.target.auto_load_trace;
+    request.spec.profiler_type     = ResolveProfilerType();
+    request.spec.profiler_path     = m_execution_cache.profiler_path;
+    request.spec.target_executable = m_config.target.executable;
+    request.spec.output_directory  = m_config.target.output_directory;
+    request.spec.working_directory = m_config.target.working_directory;
+    request.spec.profiler_argv     = m_execution_cache.argv;
+    request.spec.env_vars          = m_execution_cache.env_vars;
+    request.is_remote              = is_remote;
+    request.auto_load_trace        = m_config.target.auto_load_trace;
 #ifdef ROCPROFVIS_ENABLE_REMOTE
     request.remote_uri        = is_remote ? m_remote_uri : nullptr;
 #endif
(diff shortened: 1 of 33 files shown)


src/view/src/profiler/rocprofvis_profiler_launcher_dialog.cpp
drchen-amd, line 1192: IsAnyItemActive does not seem right. I see this being spammed each time I interact with literally anything in the app.

`Profiler tool 'rocprof-sys-run.exe' was not found in $ROCM_PATH/bin or on $PATH`

drchen-amd:
Some elements are cut off at larger font sizes regardless of available width.
image

src/view/src/rocprofvis_settings_manager.cpp
```diff
@@ -982,10 +981,9 @@ void
 SettingsManager::DeserializeProfilerSettings(jt::Json& json)
 {
     jt::Json& ps = json[JSON_KEY_GROUP_SETTINGS][JSON_KEY_SETTINGS_CATEGORY_PROFILER];
-    if(ps[JSON_KEY_SETTINGS_PROFILER_PATH].isString())
-    {
-        m_profilersettings.profiler_path = ps[JSON_KEY_SETTINGS_PROFILER_PATH].getString();
-    }
+    // A "profiler_path" key written by an older development build is read by
+    // nothing: the path is no longer part of the launch input, so a settings file
+    // cannot name the binary to execute.
```
drchen-amd: Does not seem useful.

src/view/src/profiler/rocprofvis_profiler_launcher_dialog.cpp
tomk-amd: Noted - I can look to fix this in this PR - but this is out-of-scope..

src/view/src/rocprofvis_settings_manager.cpp
tomk-amd: agreed

tomk-amd:
> Some elements are cut off at larger font sizes regardless of available width. 

Also not in the scope of this PR..

drchen-amd (approved):
I will just leave it at the UI is inconsistent, some examples: https://github.com/ROCm/roc-optiq/issues/1016.

PR #994, opened 2026-08-12
Fixed deadlock when cancelling counter track data query

### LINKED ISSUE #678: [Defect]: Tracks keep "loading" even though data has been fetched
The loading indicator remains in description area even though the track data appears to have been fetched.

https://github.com/user-attachments/assets/d2d193fd-be87-4676-acc8-752c976d676d

Happened with the sample attached to Jira: AIPROFVIS-98

When trying to close file application froze, looks like it is stuck trying to cancel these "ghost" fetch requests:

```
[2026-02-18 11:12:53.828] [rocprofvis-log] [warning] Failed to cancel request 72902018968059911: 5
2
[2026-02-18 11:12:53.828] [rocprofvis-log] [warning] FreeRequests: cancelling request 72902018968059912 of type 1
2
[2026-02-18 11:12:53.828] [rocprofvis-log] [warning] Failed to cancel request 72902018968059912: 5
2
[2026-02-18 11:12:53.829] [rocprofvis-log] [warning] FreeRequests: cancelling request 72902018968059913 of type 1
2
[2026-02-18 11:12:53.829] [rocprofvis-log] [warning] Failed to cancel request 72902018968059913: 5
2
[2026-02-18 11:12:53.829] [rocprofvis-log] [warning] FreeRequests: cancelling request 72902018968059914 of type 1
2
[2026-02-18 11:12:53.829] [rocprofvis-log] [warning] Failed to cancel request 72902018968059914: 5
````
When I pause via debugger, application seems to be trapped here:

Image

### PR DESCRIPTION
**Pull request by vstempen**
Fixed deadlock when cancelling counter track data query. Set idle request state when no data retrieved

## Motivation

Customers complain track are stuck loading data when actively zooming in/out timeline view. 

## Technical Details

There was premature exit without updating slice completion status when reading counter track from database.
The change changes the logic to complete slice when execution breaks.
The change also sets Idle status to track request state in order to be able requesting data again in case of failure.

### DIFF
--- a/src/view/src/rocprofvis_flame_track_item.cpp
+++ b/src/view/src/rocprofvis_flame_track_item.cpp
@@ -321,7 +321,6 @@ FlameTrackItem::ExtractPointsFromData()
 
     if(event_track->GetData().empty())
     {
-        m_request_state = TrackDataRequestState::kIdle;
         spdlog::debug("No data for track {}", m_track_id);
         return false;
     }
--- a/src/view/src/rocprofvis_flame_track_item.cpp
+++ b/src/view/src/rocprofvis_flame_track_item.cpp
@@ -321,6 +321,7 @@ FlameTrackItem::ExtractPointsFromData()
 
     if(event_track->GetData().empty())
     {
+        m_request_state = TrackDataRequestState::kIdle;
         spdlog::debug("No data for track {}", m_track_id);
         return false;
     }
--- a/src/view/src/rocprofvis_line_track_item.cpp
+++ b/src/view/src/rocprofvis_line_track_item.cpp
@@ -384,6 +384,7 @@ LineTrackItem::ExtractPointsFromData()
     if(sample_track->GetData().empty())
     {
         spdlog::debug("No data for track {}", m_track_id);
+        m_request_state = TrackDataRequestState::kIdle;
         return false;
     }
     const std::vector<TraceCounter>& track_data = sample_track->GetData();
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/view/src/rocprofvis_flame_track_item.cpp [hunk 1]
```diff
@@ -321,6 +321,7 @@ FlameTrackItem::ExtractPointsFromData()
 
     if(event_track->GetData().empty())
     {
+        m_request_state = TrackDataRequestState::kIdle;
```
drchen-amd: This breaks the track loading indicator. 

Before the indicator would be active until entire graph loaded in, now the indicator shows for a second before disapearing and chunks of the graph are still poping in well after. Easy to see with dense tracks that take a while to load.

vstempen: Yes, agree. That's not good fix. We need more sophisticated fix for the request state change. It doesn't take cancellation into account. It should not stay kRequesting when there is no jobs getting data for the track. Suggestions?

src/view/src/rocprofvis_line_track_item.cpp
```diff
@@ -384,6 +384,7 @@ LineTrackItem::ExtractPointsFromData()
     if(sample_track->GetData().empty())
     {
         spdlog::debug("No data for track {}", m_track_id);
+        m_request_state = TrackDataRequestState::kIdle;
         return false;
```
tomk-amd: request state should not be altered here

src/view/src/rocprofvis_flame_track_item.cpp
tomk-amd: don't change requested state here, please revert

tomk-amd (changes requested):

src/view/src/rocprofvis_flame_track_item.cpp
vstempen: reverted

tomk-amd (approved):

PR #995, opened 2026-08-12
View: Add time-range selection indicators across timeline, tables, and overview

dhingora-amd:
## Motivation

When a time-range selection is active, several views silently recompute or filter
their results, but nothing signals that on screen. This PR makes it explicit which
data depends on the current selection, and makes the selected region easy to spot.

## Technical Details

- **Pills:** per-track analysis stat pills (counter min/max/avg/stddev, queue
  utilization) tint blue while a selection is active, neutral otherwise
  (`Pill::SetRangeAffected`).
- **Tables:** the events and top-events tables show a scope indicator in the header
  — accent "Limited to time-range selection" (with a span tooltip) vs dim
  "Full trace".
- **Timeline:** tracks outside the selection are dimmed (spotlight scrim); flame
  events that only partially fall inside are repainted at full brightness above the
  scrim so they still read as included in the results.
- **Overview histogram:** bars outside the current view are greyed (minimap), with
  bars straddling a view edge split into in/out slices; start/end markers show where
  the selection sits in the full trace; the overview's time axis is aligned with the
  timeline graph when fully zoomed out.


src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -610,6 +614,31 @@ FlameTrackItem::DrawBox(ImVec2 start_position, ChartItem& chart_item, float dura
...
     draw_list->AddRectFilled(rectMin, rectMax, rectColor, rounding);
 
+    if(mark_partial)
+    {
+        // The dimming scrim sits above the tracks, so repaint the slice(s) of this
+        // event that fall outside the selection at full brightness on the foreground
+        // list (clipped to the track) - it then reads as included in the results.
+        // Only the outer corners are rounded so the internal cut stays flush with
+        // the in-selection part.
+        ImDrawList* top_layer = ImGui::GetForegroundDrawList(ImGui::GetWindowViewport());
```
drchen-amd: You cannot use this approach because it will punch through all other widgets and be topmost.

src/view/src/rocprofvis_track_item.h
```diff
@@ -72,6 +73,8 @@ class Pill
     bool                              m_show_pill_label;
     bool                              m_active;
     std::optional<size_t>             m_accent_color;
+    // Tints the value text blue to flag dependence on the time-range selection.
+    bool                              m_range_affected;
```
drchen-amd: I suggest keeping a generic text color member instead of this bespoke range flag. That way it is easier for other use cases that need to alter color to re-use.

drchen-amd (commented):
I don't like how it makes the event names hard to read.

Image

src/view/src/rocprofvis_flame_track_item.cpp
```diff
@@ -610,6 +614,31 @@ FlameTrackItem::DrawBox(ImVec2 start_position, ChartItem& chart_item, float dura
     float rounding = 2.0f;
     draw_list->AddRectFilled(rectMin, rectMax, rectColor, rounding);
 
+    if(mark_partial)
+    {
+        // The dimming scrim sits above the tracks, so repaint the slice(s) of this
+        // event that fall outside the selection at full brightness on the foreground
```
tomk-amd: this is not needed, we do not want partial highligting here

src/view/src/rocprofvis_track_item.h
```diff
@@ -72,6 +74,9 @@ class Pill
     bool                              m_show_pill_label;
     bool                              m_active;
     std::optional<size_t>             m_accent_color;
+    // Overrides the value text color when set; unset falls back to the default.
+    // Generic so any state (e.g. time-range dependence) can retint the pill.
```
drchen-amd: Line 78 should be removed. It irrelevant outside of review comments for this PR.

src/view/src/rocprofvis_multi_track_table.cpp
```diff
@@ -275,6 +276,50 @@ MultiTrackTable::Render()
...
+                                    0.0f);
+                    ImGui::TextColored(accent, "%s", ICON_CROP);
+                    ImGui::PopFont();
+                    ImGui::SameLine(0.0f, style.ItemInnerSpacing.x);
+                    ImGui::TextColored(accent, "Limited to time-range selection");
+                    if(ImGui::IsItemHovered())
+                    {
+                        SetTooltipStyled(
+                            "These results reflect the current time-range selection "
+                            "(span: %s).\nClear the selection on the timeline to see "
```
drchen-amd: `Clear the selection on the timeline to see the full trace.`

I don't think this is needed because it seems self-explanatory.

src/view/src/rocprofvis_multi_track_table.cpp
```diff
@@ -275,6 +276,50 @@ MultiTrackTable::Render()
...
+                    }
+                }
+                else
+                {
+                    ImGui::TextDisabled("Full trace");
+                    if(ImGui::IsItemHovered())
+                    {
+                        SetTooltipStyled(
+                            "No time-range selection - these results cover the full "
+                            "trace.\nDrag on the timeline ruler to limit results to a "
```
drchen-amd: `Drag on the timeline ruler to limit results to a time range.`

This is not accurate because you need to apply the modifier key. I don't think this needed anyway because it seems self-explanatory.

drchen-amd (approved):

src/view/src/rocprofvis_timeline_view.cpp
```diff
@@ -2560,75 +2634,172 @@
...
-    float x_view_end   = bars_pos.x + view_end_frac * bars_width;
-    float y0           = bars_pos.y;
-    float y1           = bars_pos.y + bars_height;
+            if(grey_bars)
+            {
+                // Grey the whole bar, then repaint the in-focus slice; this splits a
+                // bar that straddles a focus edge into coloured and grey parts.
+                draw_list->AddRectFilled(ImVec2(x0, y_bar), ImVec2(x1, y1),
+                                         grey_colors[i % 2], 1.5f);
+                float in_left  = std::max(x0, focus_left);
```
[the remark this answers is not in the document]
tomk-amd: please fix the shadow vars

tomk-amd (approved):
Please fix the shadow var issue mentioned by code ql, the it is good to go.

tomk-amd:
We will need to implement an indicator for Track Details as well (separate PR):

image

PR #996, opened 2026-08-12
View: Add setting to hide the topology sidebar row icons

Sajeeth-Wimalasuriyan:
The sidebar repeated an eye and a go-to-track button on every row, which crowded the tree on traces with many queues and counters. Both actions were already available from the row context menus, so the new Display setting drops the buttons and lets the labels sit flush against the tree. Hidden tracks stay marked by their dimmed label.

 
image
image


amokiche-amd (approved):
All looks good, nice work

PR #998, opened 2026-08-13
Fixed track stuck loding data after request cancellation

vstempen:
## Motivation

After fixing dead lock on data model side, a timeline track may still be stuck in loading animation or be empty.

1. Fixed ReleaseData to cancel pending requests before freeing track data, and only force-free when all requests are successfully cancelled — preventing deletion of partially-assembled data while chunks are still in flight
2. Fixed ReleaseData to reset m_request_state to kIdle when all pending requests are cancelled, unblocking re-requests when the track scrolls back into view
3. Fixed ExtractPointsFromData to reset m_request_state to kIdle when m_pending_requests is empty, handling the case where some in-flight requests could not be cancelled and landed after partial data was freed
4. Fixed RequestDataIfEmpty to detect and re-request tracks with incomplete data (!AllDataReady()) in addition to tracks with no data
5. Fixed CreateRawSampleData and CreateRawEventData to skip AddChunk when the request was cancelled, preventing empty or stale chunks from being committed and falsely satisfying AllDataReady()
6. Fixed request cancellation on Controller level to correctly return cancellation status instead of an out-of-range status

### DIFF
--- a/src/controller/src/system/rocprofvis_controller_track.cpp
+++ b/src/controller/src/system/rocprofvis_controller_track.cpp
@@ -182,7 +182,10 @@ rocprofvis_result_t Track::FetchSegments(double start, double end, void* user_pt
             });
         }
 
-        result = m_segments.FetchSegments(start, end, user_ptr, future, func);
+        // Fetching segnents when request cancelled will change return status to kRocProfVisResultOutOfRange.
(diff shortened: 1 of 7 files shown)


tomk-amd:
Closing superseded, by https://github.com/ROCm/roc-optiq/pull/1000.

PR #999, opened 2026-08-14
UI tests: add AIPROFVIS-253 + -296 selection/dedup tests

jeremynugara:
Adds two UI tests exercising event selection and shared-db handling on the trace view.

## Changes
- `sys_stacked_event_select_distinct` — clicks two events that overlap in time but stack on different rows, asserts the clicks select two distinct events. Motivated by AIPROFVIS-253, where overlapping stacked events shared a merged hit-region and clicks collapsed onto the first event.
- `sys_shared_db_open_dedups_and_switches` — opening a second file that shares an already-open source .db switches to the existing tab instead of opening a duplicate (AIPROFVIS-296).
- Adds a `ShowSummaryGuard` RAII helper so the Summary tests restore `show_summary` on every exit path, including SKIP early-returns.

### DIFF
--- a/src/app/test/app_tests.cpp
+++ b/src/app/test/app_tests.cpp
@@ -18,6 +18,8 @@
 #include "rocprofvis_summary_view.h"
 #include "icons/rocprovfis_icon_defines.h"
 #include <string>
+#include <filesystem>
+#include <fstream>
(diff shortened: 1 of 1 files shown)


Sajeeth-Wimalasuriyan (approved):
Don't see any issues.

PR #1002, opened 2026-08-14
UI tests: add measurement-clear and track-details-select tests

jeremynugara:
## Changes
- `sys_measurement_clear_button` -- exercises the timeline measurement toolbar Clear button (enter measure mode, place two points, clear, assert state resets and mode exits).
- `sys_track_details_populates_on_select` -- asserts the track-details panel populates when a track is selected via TimelineSelection.

### DIFF
--- a/src/app/test/app_tests.cpp
+++ b/src/app/test/app_tests.cpp
@@ -1228,16 +1228,8 @@ void RegisterAppTests(ImGuiTestEngine* e)
     t = IM_REGISTER_TEST(e, "app", "sys_track_details_populates_on_select");
     t->TestFunc = [](ImGuiTestContext* ctx)
     {
-        AppWindow* app = AppWindow::GetInstance();
-        Project* project = app->GetCurrentProject();
-        IM_CHECK(project != nullptr);
-        if (project == nullptr) return;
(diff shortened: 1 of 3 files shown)


src/app/test/app_tests.cpp
```diff
@@ -1225,6 +1225,70 @@ void RegisterAppTests(ImGuiTestEngine* e)
...
     };
 
+    t = IM_REGISTER_TEST(e, "app", "sys_track_details_populates_on_select");
+    t->TestFunc = [](ImGuiTestContext* ctx)
+    {
+        AppWindow* app = AppWindow::GetInstance();
+        Project* project = app->GetCurrentProject();
+        IM_CHECK(project != nullptr);
+        if (project == nullptr) return;
+        TraceView* tv = dynamic_cast<TraceView*>(project->GetView().get());
```
tomk-amd: use:
        
TraceView* tv = GetTraceViewOrSkip(ctx);

instead?

tomk-amd (approved):

src/app/test/app_tests.cpp
jeremynugara: Thanks, switched it

PR #1009, opened 2026-08-18
Update PC sampling to the new schema

amokiche-amd:
## Motivation
image
Updated PC-sampling database queries for the new UUID-based schema, including code objects, kernel symbols, instructions, source mappings, sample states, stalls, and lookup tables.
Split the monolithic PC-sampling request into three asynchronous layers:- ISA data
- Source data
- Stall/sample metadata

Reworked controller storage and properties to match the new schema.
Added per-layer synchronization so completed ISA/source/stall data is not blocked by unrelated PC-sampling work.
Added cancellation, queued replacement requests, and generation validation to safely handle rapid workload/kernel/source switching.
Refactored the view to load ISA immediately while fetching source and stall data only when requested.
Renamed ComputeCodeView and its files to ComputeIsaView in [rocprofvis_compute_isa_view.cpp](C:/ROCm/compute/roc-optiq/src/view/src/compute/rocprofvis_compute_isa_view.cpp).
Reorganized UI model types around code objects, kernel symbols, instruction lines, source mappings, and PC sample states.
Updated controller, database, and UI architecture documentation.
Removed prototype/generated tests from this PR.

## Submission Checklist


.agents/CONTROLLER.md
```diff
@@ -163,6 +163,17 @@ rocprofvis_controller_metric_fetch_async(...);                   // compute metr
...
 ```
 
+PC-sampling data has one compatibility fetch plus three stage-specific
+fetchers. Code View uses the stage-specific APIs so opening optional UI does
+not block the initial ISA display:
+
+```c
+rocprofvis_controller_pc_sampling_fetch_mandatorys_async(...); // ISA dependencies + lines
+rocprofvis_controller_pc_sampling_fetch_source_async(...);     // source metadata/correlation/lines
+rocprofvis_controller_pc_sampling_fetch_stalls_async(...);     // states/reasons/instruction samples
```
tomk-amd: rocprofvis_controller_pc_sampling_fetch_mandatorys_async  mandotorys is not a proper word, maybe change this to rocprofvis_controller_pc_sampling_fetch_isa_lines_async or rocprofvis_controller_pc_sampling_fetch_isa_dependencies_async

src/view/src/compute/rocprofvis_compute_isa_view.h
```diff
@@ -0,0 +1,201 @@
...
+class IsaCodeWidget;
+class DataProvider;
+enum class PcSamplingLayer : uint32_t;
+
+struct LineSelection
+{
+    uint64_t hovered_line  = 0;
+    uint64_t selected_line = 0;
+};
+
```
tomk-amd, line 38: how are lines set to  un-hovered or un-selected?  is zero the INVALID_INDEX in this case?

Image

But I don't think it can be, since src lines can start from 0, unless that is a bug.  (Strange that ISA starts 1 and src at 0)..

tomk-amd: Also I don't see how `hovered_line` is ever un-set.. once you hover over something it will stay hovered for ever...

For example move mouse to hover a line, then move off of the src line window, lines stay hovered for ever.

.agents/CONTROLLER.md
amokiche-amd: agree, already resolved

src/view/src/compute/rocprofvis_compute_isa_view.h
amokiche-amd: yep its a bug, I am in progress to fix it

amokiche-amd:
I got a little carried away and added a small feature while fixing this issue. Now, if you select a line, the opposite window will scroll to the corresponding location. If you select lines related to a different source file, the desired file will be selected and the corresponding source code will be displayed.

src/view/src/compute/rocprofvis_compute_isa_view.cpp
```diff
@@ -0,0 +1,949 @@
...
+
+    ImGui::EndChild();
+}
+
+void
+ComputeIsaView::RenderSourceFileDropdown()
+{
+    constexpr const float DROPDAWN_SIZE = 300.0f;
+    if(!m_source_layout_item->m_visible || m_source.files.empty()) return;
+
```
tomk-amd: Typo:  DROPDAWN_SIZE -> DROPDOWN_SIZE

src/controller/src/compute/rocprofvis_controller_pc_sampling.cpp
```diff
@@ -14,9 +14,38 @@ PcSampling::PcSampling()
...
+    if(in_range(kRPVControllerPCSamplingNumCodeObjects,
+                kRPVControllerPCSamplingInstructionLineInstruction))
+        return m_isa_data_mutex;
+
+    if(in_range(kRPVControllerPCSamplingNumSourceFiles,
+                kRPVControllerPCSamplingSourceLineContent) ||
+       in_range(kRPVControllerPCSamplingNumInstructionSourceLines,
+                kRPVControllerPCSamplingInstructionSourceLineFrameIndex) ||
+       property == kRPVControllerPCSamplingInstructionSourceLineSourceFileUuid)
+        return m_source_data_mutex;
```
tomk-amd: This is clever - but potentially dangerous if future maintainers add to the enum without considering this code relies on ordering, maybe add a strong note in the enum that the mutex returned depends on ordering..

tomk-amd (approved):

tomk-amd:
the "permanent" hover issue still exists - but this can be fixed in future PR when the UI interface gets additional treatment

PR #1022, opened 2026-08-21
Reconcile test_main.cpp with main.cpp

jeremynugara:
`test_main.cpp` was forked from an old `main.cpp` and had drifted. Updates the four spots that were out of date: modifier handling (#937), macOS Vulkan setup (#883), log path (#886), and idle wait (#913).


tomk-amd (approved):
Not sure if test main ultimately needs to shadow the real main in functionality completely.. the automated tests rely on ImGUI solely, I don't think they exercise any of the glfw / vulkan / opengl functionality?

jeremynugara:
> Not sure if test main ultimately needs to shadow the real main in functionality completely.. the automated tests rely on ImGUI solely, I don't think they exercise any of the glfw / vulkan / opengl functionality?

That's true. I can keep the main -> test_main updates limited to changes that might impact ImGui.

PR #1024, opened 2026-08-24
Compare: union A/B group-by columns in the shared combo

tomk-amd:
Compare: union A/B group-by columns in the shared combo

Show every eligible column from either source, tag A-only and B-only names, and skip grouping on a pane that does not have the chosen column. Non-compare tables still use their own header.


Sajeeth-Wimalasuriyan (approved):
seems alright

PR #1026, opened 2026-08-24
Profiler hub interface library draft

vstempen:
## Motivation

This PR is just to save and share the work done for Profiler Hub interface with Optiq.
It's a bridge library. It should speed up profiler-hub integration and make it functional.
The library code should be a part of profiler hub repository in future, if design is accepted.
We need to decide how profiler hub will be integrated into a client. One way is by just adding headers and library binary.
Another way integrating with "git add submodule" and "git sparse-checkout" from rocm-systems.
I do not currently have permissions to push it to rocm-systems branch, so I created temporary branch in roc-Optiq repository for profiler-hub interface library.
Proposing integrating profiler hub into roc-Optiq in 3 stages:
First stage of integration is creating a test utility, which will run profiler-hub and data-model in parallel and check functionality, data integrity and performance.
Second stage is using profiler hub as one of data-model adapters. 
Third stage is completely move to profiler hub with data-model as thin interface layer between profiler hub and controller.

## Technical Details

Profiler hub code is not a part of this change. We need to decide how do we integrate it first.
The interface library is to be used with profiler-hub test utility developed within roc-optiq repository as a first stage of integration.
The interface library is located in thirdparty/profiler-hub-client/optiq directory. Cmake file will require thirdparty/profiler-hub to build the library.
It's just a draft and main purpose is to speed up changes in profiler-hub. It also serves as a source of requirements.
All the required changes are listed in either missing_t class or  profiler_hub::missing_types namespace. 
The requirements may not be 100% complete at this point, but we need to set a direction and establish integration process first.

### DIFF
--- a/thirdparty/profiler-hub-client/optiq/profiler_hub_interface_types.h
+++ b/thirdparty/profiler-hub-client/optiq/profiler_hub_interface_types.h
@@ -95,3 +95,12 @@ typedef enum profiler_hub_flow_direction_t
     kProfilerHubDirectionOutgoing,
     kProfilerHubDirectionIncoming,
 } profiler_hub_flow_direction_t;    
+
+// Database type
+typedef enum profiler_hub_db_type_t {
+    // not supported by profiler hub
+    kDbNotSupported, 
+    // supported by profiler hub
+    kDbSupported,
+
+} profiler_hub_db_type_t;
--- a/thirdparty/profiler-hub-client/optiq/profiler_hub_interface_types.h
+++ b/thirdparty/profiler-hub-client/optiq/profiler_hub_interface_types.h
@@ -1,0 +1,97 @@
+#pragma once
+
+#include <cstdint>
+#include <variant>
+#include <profiler-hub/reader_types.hpp>
+
+typedef void* profiler_hub_handle_t;
+typedef profiler_hub_handle_t profiler_hub_trace_handle_t;
+typedef profiler_hub_handle_t client_trace_handle_t;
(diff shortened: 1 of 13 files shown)


thirdparty/profiler-hub-client/optiq/profiler_hub_interface_types.h
```diff
@@ -0,0 +1,106 @@
...
+typedef profiler_hub_handle_t profiler_hub_trace_handle_t;
+typedef profiler_hub_handle_t client_trace_handle_t;
+typedef profiler_hub_handle_t profiler_hub_future_handle_t;
+typedef profiler_hub_handle_t profiler_hub_value_handle_t;
+typedef profiler_hub_handle_t profiler_hub_timeslice_handle_t;
+typedef profiler_hub_handle_t profiler_hub_flowtrace_handle_t;
+typedef profiler_hub_handle_t profiler_hub_ext_data_handle_t;
+typedef profiler_hub_handle_t profiler_hub_table_handle_t;
+typedef profiler_hub_handle_t profiler_hub_table_row_handle_t;
+typedef profiler_hub_handle_t profiler_hub_call_stack_handle_t;
```
adjordje-amd, line 17: Going with `void*` here pretty much kills the types

vstempen: Yeah, that's the goal. All the handlers are void* type. Basically, it's C interface feature, or even ABI feature. The types created out of void* are just for better readability and for placing data into proper container. I'm still not convinced that C++ API interface is better. But We can change the types when we agree on interface.

PR #1027, opened 2026-08-25
Simplified Table Filter

drchen-amd:
Add simplified UI for filtering paginated tables that abstracts SQL concepts. Existing controls still accessible under "Advanced" mode. Currently applicable to event and sample tables as others (search, top events, summary) ignore filtering parameter in backend, but easy to expand to them otherwise.

### DM
- `TableProcessor`
  - Defined column data types for public columns.
  - Corrected output of an invalid filter claiming to have identical row count to that of unfiltered result despite output contents being empty.

### Controller
- Added `Array` property `kRPVControllerArrayEntryTypeIndexed` for getting the data type of each entry.
- Populated `kRPVControllerTableColumnTypeIndexed` for `SystemTable`.
- Factored out common getters out of derived `Table`s into base `Table`.

### View
- Added column type to table models.
- `InfiniteScrollTable`
  - Consolidated data fetch, cancellation, deferral, into singe entry point.
    - Derives are responsible for a `UpdateFetchParams` implementation.
    - When data fetch is needed for any reason, signal request via private flag or public function call. Fetch sequence will pick up all parameters via `UpdateFetchParams` and handle the rest behind the scenes.
  - Added simplified filtering controls.
- `MultiTrackTable `
  - Reworked existing filtering controls. Previous implementation used nonstandard dimensions (ex. 
`ImGui::PushStyleVar(ImGuiStyleVar_CellPadding, ImVec2(style.FramePadding.x, style.ItemSpacing.y * 0.35f))`, making it finicky to align new additions.
  - Updated for `InfiniteScrollTable` changes.
  - Added additional filtering controls.
- `KernelInstanceTable`, `EventSearchTable`
  - Updated for `InfiniteScrollTable` changes.
- Switched `TextInputWithClear` to use `InputTextStringWithHint`
- Switched `ICON_COPY` to non-outline version to be more consistent wiht other icons.

### DIFF
--- a/src/app/test/app_tests.cpp
+++ b/src/app/test/app_tests.cpp
@@ -1369,10 +1369,7 @@ void RegisterAppTests(ImGuiTestEngine* e)
         // (Launch/Dispatch/MemoryCopy/MemoryAllocate/LaunchSample -- see
         // EventSearch::Search). If the sample db changes, update it to a term the
         // new db contains.
-        char* buf = es->TextInput();
-        IM_CHECK(buf != nullptr);
-        if (buf == nullptr) return;
-        snprintf(buf, es->TextInputLimit(), "%s", "hipLaunchKernel");
+        es->TextInput() = "hipLaunchKernel";
         es->Search();
         ctx->Yield(2);
         IM_CHECK(es->Searched() == true);
--- a/src/controller/inc/rocprofvis_controller_enums.h
+++ b/src/controller/inc/rocprofvis_controller_enums.h
@@ -598,6 +598,8 @@ typedef enum rocprofvis_controller_array_properties_t : uint32_t
     kRPVControllerArrayNumEntries = __kRPVControllerArrayPropertiesFirst,
     // Indexed entry.
     kRPVControllerArrayEntryIndexed,
+    // Primitive type of the indexed entry.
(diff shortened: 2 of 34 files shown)


tomk-amd:
suggestion:

for columns that contain time values put a note in the tooltip that the values must be compared in ns as other units will not filter properly:

image

tomk-amd:
another observation, when columns are narrow the (x) button disappears (gets clipped)..  not sure what a good solution is in this case? show it on hover? or set a minimum column width (assuming this works with auto-sizing columns - i remember that getting control of column sizes with auto sizing was tricky)

image

drchen-amd:
> another observation, when columns are narrow the (x) button disappears (gets clipped).. not sure what a good solution is in this case? show it on hover? or set a minimum column width (assuming this works with auto-sizing columns - i remember that getting control of column sizes with auto sizing was tricky)
> 
> image

Yes, I have it setup such that clear only shows if available width is at least `2x` the clear button's width. Otherwise, what little space you have is either entirely covered or dominated by the clear button and I figured the text input itself is more important to see. I don't think show on hover would help this scenario. The clear button is only a nice to have in my opinion since you can always clear the text yourself via keyboard, especially since now it will auto highlight on click.

tomk-amd (approved):

PR #1032, opened 2026-08-31
test: add UI coverage for event search, table, and details views

jeremynugara:
Adds six headless UI tests covering the event search box, event table, and event details panel.

- sys_event_search_zero_result_and_clear: searches a term that matches nothing,  asserts zero results, and checks clear resets the searched state.
- sys_event_search_multi_substring: runs a quoted multi-term query and asserts it returns results.
- sys_event_table_updates_on_track_deselect: deselects one of several selected tracks and asserts the table refetches and drops that track's rows (AIPROFVIS-117).
- sys_event_table_aggregate_changes_rows: picks a group-by column, submits, and asserts the row count changes as events collapse into groups.
- sys_event_details_shows_hip_args: selects a HIP API event and checks its argument list shows in the details panel (AIPROFVIS-81).
- sys_project_missing_source_db_error: opens a .rpv whose trace is gone, asserts it fails with a message naming the trace and leaves no file behind (AIPROFVIS-297).

### DIFF
--- a/src/app/test/app_tests.cpp
+++ b/src/app/test/app_tests.cpp
@@ -2644,4 +2644,125 @@ void RegisterAppTests(ImGuiTestEngine* e)
         IM_CHECK(message.find(expected_path) != std::string::npos);
         IM_CHECK(still_missing);
     };
+
+    // AIPROFVIS-81: Event Details dropped the argument list for HIP API events.
+    // Presence/shape only -- the args come back asynchronously through the
+    // controller, with no single table to build a value oracle from.
+    t = IM_REGISTER_TEST(e, "app", "sys_event_details_shows_hip_args");
+    t->TestFunc = [](ImGuiTestContext* ctx)
+    {
+        TraceView* tv = GetTraceViewOrSkip(ctx);
+        if (!tv) return;
+        AnalysisView* av = TraceViewTestPeer{*tv}.AnalysisViewPtr();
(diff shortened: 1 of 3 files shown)


src/app/test/app_tests.cpp
```diff
@@ -2148,4 +2219,550 @@ void RegisterAppTests(ImGuiTestEngine* e)
...
+        ctx->ItemInput("**/search_bar/##input_text_with_clear");
+        ctx->KeyCharsReplaceEnter("\"hip\"\"Launch\"");
+        ctx->Yield(2);
+        IM_CHECK(es->Searched() == true);
+
+        // The fetch is deferred. Let it drain (Update re-runs Search when the
+        // request completes) before reading the result count.
+        for (int i = 0; i < 60 && EventSearchTestPeer{*es}.RequestPending(); i++) ctx->Yield(2);
+        ctx->Yield(5);
+        IM_CHECK(EventSearchTestPeer{*es}.ResultCount() > 0);
```
dhingora-amd: Rather than just checking if the count is greater than 0, ensure that the count is strictly less than just searching "hip" (it should be strictly less since hipMalloc)

Sajeeth-Wimalasuriyan (commented):
I am ok with this Dipal found a potential weak point in testing please wait for his review before merging

Sajeeth-Wimalasuriyan (approved):
its good been sitting long enough

PR #1033, opened 2026-09-01
MemoryManager Improvements

drchen-amd:
Addresses following issues observed while navigating the timeline:
1.  Intermittent crash in `MemoryManager::ManageLRU`
    - Acquire lru lock to prevent `std::sort` from running while contents were being manipulated in other threads.
2.  Intermittent crash in `MemoryManager::Delete`
    - Acquire pool lock to prevent it and `NewEvent`, `NewSample`, `NewSampleLod`, `Cleanup` from modifying pool simultaneously.
3.  Memory usage growing over time
    - Added deletion of events not stored in a segment. `Segment::Insert` will not store an event if it already holds an event with the same timestamp. These events were invisible to `MemoryManager` and could not be evicted. 
    - Registered all fetched segments with `MemoryManager`. Only segments that are read in `Segment::Fetch` were registered with `MemoryManager` (`MemoryManager::AddLRUReference`). If fetch was cancelled, there would be skipped segments in aforementioned function who are never registered with `MemoryManager` and cannot be evicted.
    - Replaced array identifier in `LRUMember` with `uint64_t id` instead of array's ptr. It is possible for a newly allocated array to reuse address of an array that had already be freed. This would cause `LRUMembers` of the already freed array to be considered in use.

Tested using automated test that constantly pans and zooms timeline and switches between ten tabs for 46 hours and counting.
| Time Window| Beginning Total App Memory Usage (MB) | End Total App Memory Usage (MB) | Rate (MB/hr) |
|---|---|---|---|
| last 36h | 9,496 | 11,066 | +43.6 |
| last 24h | 10,259 | 11,066 | +33.6 |
| last 18h | 10,388 | 11,066 | +37.7 |
| last 12h | 10,584 | 11,066 | +40.2 |
| last 8h | 10,981 | 11,066 | +10.6 |
| last 6h | 10,996 | 11,066 | +11.7 |
| last 4h | 11,010 | 11,066 | +14.0 |

(For reference, unmodified app would trigger Ubuntu OOM killer in ~22 hours on 128 GB system)

### DIFF
--- a/src/controller/src/rocprofvis_controller.cpp
+++ b/src/controller/src/rocprofvis_controller.cpp
@@ -714,11 +714,14 @@ void rocprofvis_controller_array_free(rocprofvis_controller_array_t* object)
     RocProfVis::Controller::ArrayRef array(object);
     if (array.IsValid())
     {
-        if(array.Get()->GetContext() &&
-           ((RocProfVis::Controller::SystemTrace*) array.Get()->GetContext())->GetMemoryManager())
+        if(array.Get()->GetContext())
         {
-            ((RocProfVis::Controller::SystemTrace*)array.Get()->GetContext())->GetMemoryManager()->CancelArrayOwnership(&array.Get()->GetVector(),
-                                      RocProfVis::Controller::kRocProfVisOwnerTypeGraph);
+            RocProfVis::Controller::MemoryManager* mgr = ((RocProfVis::Controller::SystemTrace*) array.Get()->GetContext())->GetMemoryManager();
+            if(mgr)
+            {
+                mgr->CancelArrayOwnership(array.Get()->GetArrayId(), RocProfVis::Controller::kRocProfVisOwnerTypeTrack);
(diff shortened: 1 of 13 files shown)


tomk-amd (approved):

drchen-amd:
New 118 hour run:
| Time Window | Beginning Total App Memory Usage (MB) | End Total App Memory Usage (MB) | Rate (MB/hr) |
| --- | --- | --- | --- |
| last 96h | 14,094 | 15,488 | +14.5 |
| last 72h | 14,682 | 15,488 | +11.2 |
| last 48h | 15,343 | 15,488 | +3.0 |
| last 36h | 15,391 | 15,488 | +2.7 |
| last 24h | 15,447 | 15,488 | +1.7 |
| last 18h | 15,456 | 15,488 | +1.8 |
| last 12h | 15,461 | 15,488 | +2.2 |
| last 8h | 15,461 | 15,488 | +3.4 |
| last 6h | 15,453 | 15,488 | +5.8 |
| last 4h | 15,468 | 15,488 | +5.0 |

PR #1034, opened 2026-09-01
Refactor (re-write) UI topology logic

tomk-amd:
## Motivation

Simplify the topology logic / model on the UI side.  Remove track topology object.

Fix the issue where node ordering can differ on different systems. (stl hash map implementation differences).

AI descriptions follow:

The controller already hands the View a parent-linked topology tree. The View
flattened it into eight `unordered_map`s with the parent/child edges degraded to
id vectors, and then `TrackTopology` reassembled a nested mirror hierarchy
(`TopologyModel` and its `*Model` structs) back out of those maps in ~340 lines
of five-deep loops. Everything downstream navigated the mirror through LUT
chains like `node_lut[n]->processor_lut[p]->queue_lut[id]`.

## Technical Details

This replaces both layers with one tree that the UI reads directly.

- **`TopologyDataModel` becomes `TopologyTree`** (`src/view/src/model/`): an
  arena owning every node, with `TopologyNode` as the base for the `*Info`
  structs. Children are real edges, so the `*_ids` vectors are gone, and a queue
  exists once while hanging off both its processor and each stream that
  dispatches to it. `DeviceInfo` is renamed `ProcessorInfo` to match the
  controller.
- **Nodes bind `track_id`, not `graph_index`.** A graph index is presentation
  order, which now changes on every sort, and caching it in the topology created
  a cycle: tree order determined `TrackInfo::index`, while `UpdateGraphs()` wrote
  `index` back into the tree. A drag-reorder fired `kTrackMetadataChanged`, which
  made `TrackTopology` recompute the very order vector `TimelineView` was
  pointing at. The tree is now a pure input to ordering and never a consumer.
- **`TrackTopology` is deleted, not merged.** Its two remaining jobs went to
  their owners: `SideBar` builds the presentation tree it draws, and the sort
  order lives in `TopologyTree::GetTrackOrder()`. `TreeNode` / `LeafNode` keep
  only the genuinely presentational state (group headers, eye-state cache, node
  colors, inline processor rows), so the same conflation cannot reappear.
- **`TrackDetails` builds its own tables for the selected tracks.** `InfoTable`
  (renamed `DetailsTable` to stop colliding with the database type of that name)
  was precomputed for every node in the trace and `FormatCells()` re-walked the
  entire topology on every time-format change — to serve a pane showing one to
  five tracks. Cells now carry an `is_time` flag instead of a `std::function`
  capturing `TrackTopology* this` inside the data model.
- **`GetRevision()` replaces the shared `Dirty()` flag and `kTopologyChanged`,**
  so each consumer remembers the revision it last built against and the sidebar
  rebuilding no longer forces Track Details to rebuild.

### Fixes along the way

- **Stream subtrees came up empty on multinode and compare traces.** A stream's
  processors were looked up by `id.fields.id` while the processor map was keyed
  by `id.value`. The full `TopologyId` is now the node's id.
- **"Sort by topology" silently fell back to Default** on any trace with tracks
  the sidebar tree did not cover, because `SetTrackIndex` rejects anything short
  of a full permutation. `TimelineView` now appends the uncovered tracks.
- **`GetNodeDisplayIndex()` copied every node id into a vector and sorted it on
  each call,** once per track during render. The rank is assigned once at load.

Net -294 lines.

## Test plan

- [x] Both UI suites pass (`ctest -R roc-optiq-ui-tests`), at each of the three
      commits rather than only at the tip.
- [x] Added `sys_sidebar_topology_tree_populates`. The topology pane had **no**
      coverage, so a projection that never got built would have rendered an empty
      sidebar and still shown green. It asserts the tree is built, every leaf's
      track id resolves to a live track, and the derived sort order is a
      complete, duplicate-free permutation. Verified the test actually executes
      by temporarily inverting its leaf-count assertion and confirming the suite
      fails.
- [x] Moved sidebar builders diffed line-by-line against the deleted originals,
      and the flattened track order reproduces the old sidebar-derived walk
      (including the inline queues under a stream leaf).
- [x] **Needs a reviewer with a multinode or compare trace.** Node ordering, the
      color-wheel index, and the `TopologyId::instance` bits all interact with
      `display_index`, and the stream-subtree keying fix above is only observable
      there. The sample traces in `sample/` are single-node.


src/view/src/model/rocprofvis_topology_model.h
```diff
@@ -5,106 +5,325 @@
...
+    // Lookup shortcuts.
+    const NodeInfo*      GetNode(uint64_t node_id) const;
+    const ProcessorInfo* GetProcessor(uint64_t processor_id) const;
+    const ProcessInfo*   GetProcess(uint64_t process_id) const;
+    const QueueInfo*     GetQueue(uint64_t queue_id, uint64_t processor_id) const;
+    const CounterInfo*   GetCounter(uint64_t counter_id) const;
+    const ThreadInfo*    GetThread(uint64_t thread_id, ThreadInfo::Kind kind) const;
+
+    ProcessorInfo* GetProcessorMutable(uint64_t processor_id) const;
+    QueueInfo*     GetQueueMutable(uint64_t queue_id, uint64_t processor_id) const;
```
drchen-amd: I don't think these GetXMutable should be const.

src/view/src/model/rocprofvis_topology_model.h
```diff
@@ -5,106 +5,325 @@
...
+    static const std::vector<TopologyNode*>& EmptyChildList();
+
+    TopologyNodeType                                       m_type;
+    uint64_t                                               m_id;
+    std::string                                            m_name;
+    std::string                                            m_header;
+    TopologyNode*                                          m_parent;
+    std::map<TopologyNodeType, std::vector<TopologyNode*>> m_children;
+    std::map<TopologyNodeType, std::vector<TopologyNode*>> m_linked_children;
+    std::vector<TopologyNode*>                             m_secondary_parents;
```
drchen-amd, line 93: I wonder if it would be better if `m_secondary_parents` and `m_linked_children` became properties of `SteamInfo/QueueInfo` instead of base? 

I don't think they're relevant for other types. `m_secondary_parents/GetSecondaryParents()` is unused at the moment, and `m_linked_children/GetLinkedChildren()` is used in two places that already know they are dealing with `StreamInfo`s. If they get moved into `StreamInfo/QueueInfo`, it would be clearer that they represent child queues/parent streams.

tomk-amd: I'd prefer for them to be general properties.  Right now queues and streams have this secondary relationship, but in the future we may encounter other cases.  I'd prefer it to be a general property of the "tree" than just some specific node types.

src/view/src/model/rocprofvis_model_types.h
```diff
@@ -154,77 +154,8 @@ struct EventInfo
...
-
-struct CounterInfo : public IterableInfo
-{
-    uint64_t    device_id;  // ID of owning device.
-    std::string description;
-    std::string units;
-    std::string value_type;
-};
+// Topology types (NodeInfo, ProcessorInfo, ProcessInfo, ...) live with the tree
+// that owns them, in rocprofvis_topology_model.h.
```
drchen-amd: I don't think knowing that something used to be here is useful.

drchen-amd (approved):

src/view/src/model/rocprofvis_model_types.h
tomk-amd: It was obviously useful to the author :)

PR #1037, opened 2026-09-03
View: Limit and auto-fit table column widths (#1023)

### LINKED ISSUE #1023: [Defect]: Limit column size in tables
Limit column width in tables. Some content, for example kernel names, can be extreme long causing auto-sized columns to grow very wide.

Image

and restore width(s) or fit them better to content after a long string has been removed

Image

dhingora-amd:
## Motivation

Closes #1023.

Table columns sized themselves to their raw content, so a single very long
value (e.g. a demangled kernel name, which can be multiple KB) would stretch a
column to thousands of pixels and force endless horizontal scrolling, making
tables hard to read. The goal is to keep columns to a sensible width, elide
overflowing text while keeping the full value accessible, and auto-fit columns
to what's on screen without fighting the user's manual resizes.

## Technical Details

- **Clamped auto-fit:** columns are fitted to the widest loaded value, clamped
  to `[MIN_COLUMN_WIDTH_EM, MAX_COLUMN_FIT_WIDTH_EM]` (font-size multiples, so it
  scales with DPI). The min keeps the filter row usable; the max stops runaway
  columns.
- **Fit timing:** a full fit runs on content changes (track selection / filter);
  scroll page-ins only ever grow columns (never shrink mid-browse); sorting and
  plain scrolling don't refit.
- **Respects manual resizing:** manual column drags are detected and skipped by
  the auto-fit. Sizing state is keyed by column name, so adding or removing
  tracks (and their columns) no longer resets widths you set yourself; only a
  full clear (all tracks deselected) restores default sizing.
- **Text elision:** cell values that exceed the column width are elided with an
  ellipsis; the full value remains available in the hover tooltip and in copy
  actions. Elision is centralized in a shared `gui_helpers` helper.
- **Filter hints:** replaced the verbose per-column placeholder strings with a
  simple "Filter" placeholder plus a hover tooltip that shows the supported
  operators (`> < = >= <= !=`) and an example, styled to match existing tooltips.

### DIFF
--- a/src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
+++ b/src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
@@ -11,6 +11,7 @@
 #include "spdlog/spdlog.h"
 #include "widgets/rocprofvis_gui_helpers.h"
 #include "widgets/rocprofvis_notification_manager.h"
+#include "imgui_internal.h"
 #include <algorithm>
 #include <sstream>
 
@@ -28,6 +29,9 @@ constexpr const char* FILTER_TEXT_HINT_STR       = "Filter: hipLaunchKernel";
 constexpr const char* FILTER_TEXT_HINT_NUMERICAL = "Filter: {>, <, =, >=, <=, !=} 30";
 constexpr const char* FILTER_TEXT_HINT_TIME =
     "Filter: {>, <, =, >=, <=, !=} {Nanoseconds}";
+
+// Cap on a column's fitted width, in multiples of the current font size.
+constexpr float MAX_COLUMN_FIT_WIDTH_EM = 40.0f;
 
 InfiniteScrollTable::InfiniteScrollTable(
     DataProvider& dp, TableType table_type,
@@ -62,6 +66,9 @@ InfiniteScrollTable::InfiniteScrollTable(
 , m_update_filter_row(false)
 , m_reset_filter_row(false)
 , m_data_changed(true)
+, m_refit_pending(false)
+, m_refit_requested(false)
+, m_columns_emptied(false)
 , m_filter_requested(false)
 , m_fetch_data(false)
 , m_fetch_cancelled(false)
@@ -248,6 +255,17 @@ InfiniteScrollTable::HandleNewTableData(std::shared_ptr<RocEvent> e)
         m_data_changed      = true;
         m_update_filter_row = m_display_filter_row;
         IndexColumns();
+
+        // Re-fit columns only for a content change (select/filter), not a sort or
+        // scroll page - else sorting would fit to whatever loaded at the top.
+        std::shared_ptr<TableDataEvent> table_event =
+            std::dynamic_pointer_cast<TableDataEvent>(e);
+        if(table_event && table_event->GetRequestID() == m_request_id &&
+           m_refit_requested)
+        {
+            m_refit_pending   = true;
+            m_refit_requested = false;
+        }
     }
 }
 
@@ -290,6 +308,12 @@ InfiniteScrollTable::Render()
     {
         m_skip_data_fetch      = true;
         m_last_total_row_count = total_row_count;
+    }
+
+    // Emptying the table forgets manual sizes so the next content re-fits fresh.
+    if(total_row_count == 0)
+    {
+        m_columns_emptied = true;
     }
 
     uint64_t row_count = 0;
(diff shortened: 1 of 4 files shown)


src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
```diff
@@ -14,6 +14,7 @@
 #include "spdlog/spdlog.h"
 #include "widgets/rocprofvis_gui_helpers.h"
 #include "widgets/rocprofvis_notification_manager.h"
+#include "imgui_internal.h"
```
drchen-amd: @tomk-amd Is this allowed?

tomk-amd: I'd prefer to not use it, but the precedent has already been set in the compare functionality PR.. The rule I'd like to adhere to is avoid if possible.  Allow if the work around introduces an unreasonable amount of complexity.

tomk-amd:
image

Please limit the width of the tool tip to something reasonable, there are examples of this being done in other areas of the UI for example on the Kernel Details tab of the Compute view:

image

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
tomk-amd: However it is included **twice** in the file please remove one of the statements

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
```diff
@@ -827,6 +978,17 @@ InfiniteScrollTable::RenderCell(const std::string* cell_text, int row, int colum
...
+    if(is_elided && ImGui::IsItemHovered())
+    {
+        ImGui::SetNextWindowSizeConstraints(ImVec2(0, 0),
+                                            ImVec2(TOOLTIP_MAX_WIDTH, FLT_MAX));
+        BeginTooltipStyled();
+        ImGui::PushTextWrapPos(ImGui::GetCursorPosX() + TOOLTIP_MAX_WIDTH);
+        ImGui::TextUnformatted(cell_text->c_str());
+        ImGui::PopTextWrapPos();
+        EndTooltipStyled();
+    }
```
drchen-amd: For time columns, up one level from this is another tooltip that just shows it in ns. When a time column is cut off, that existing one seems to take affect instead of this one. I don't know which one we would rather have, so just FYI.

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
```diff
@@ -333,6 +314,23 @@
         m_data_changed      = true;
         m_update_filter_row = m_display_filter_row;
         IndexColumns();
+
+        // A content change (select/filter) re-fits the columns fresh; any other
+        // page-in of this table's data just lets columns grow to newly loaded values.
+        std::shared_ptr<TableDataEvent> table_event =
```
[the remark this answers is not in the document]
drchen-amd: ^

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
```diff
@@ -815,10 +840,136 @@ InfiniteScrollTable::FetchData()
...
+        if(it == m_column_fit_widths.end())
+        {
+            continue;  // No auto-fit applied yet; nothing to compare against.
+        }
+        // WidthRequest only diverges from our applied value on a user drag.
+        const float delta = table->Columns[c].WidthRequest - it->second;
+        if(delta > 0.5f || delta < -0.5f)
+        {
+            m_user_sized_columns.insert(name);
+        }
```
drchen-amd: This will capture a font size or time format change and include those as user sized as well, so you lose auto fit for affected columns.

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
drchen-amd: For time columns, up one level from this is another tooltip that just shows it in ns. When a time column is cut off, that existing one seems to take affect instead of this one. I don't know which one we would rather have, so just FYI.

src/view/src/widgets/rocprofvis_infinite_scroll_table.cpp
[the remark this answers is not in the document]
drchen-amd: ^

Sajeeth-Wimalasuriyan (approved):
some nitpicks 
1. Fit runs after TableHeadersRow(), so the new WidthRequest lands next frame. First frame after a load can still flash ImGui’s uncapped auto-fit. Moving detect/fit to after TableSetupColumn and before headers would apply the cap the same frame.
2. FitColumnsToContent measures every cached row (chunk size 1000) with CalcTextSize on the full string. That is the right fit, but huge demangled names could hitch on page-in. Worth a quick scroll test on a busy kernel table, not a rewrite.
3. Time-format change reformats cells but does not re-fit. Longer timestamps elide until the next content change. Intentional given the LastResizedColumn fix; a grow on kTimeFormatChanged would be nicer.
4. ElideWithEllipsis remaining == begin then begin + 1 can split a multibyte UTF-8 character. Same class of issue as the old pop_back() loop; table data is almost all ASCII.

PR #1038, opened 2026-09-03
Improve profiler launching backend (foundations for multi-step runs)

tomk-amd:
## Motivation

Continue improving the profiling backend.  This PR adds additional pieces needed for running multiple tools.

- Fix "single shot" instrument mode.
- Started implementation of a multi stage pipeline.  This will allow multi step runs (running multiple binaries in sequence, where output of one can feed into the next).
- Add an improved output scraper, that can be configured based on the tool.

No attention given to UI unless necessary for functionality.  (no aesthetic or layout fixes and improvements).

## Technical Details 

AI generated overview:

Profiler: scrape tool output, and run a config's stages as a sequence

Phase 1 of the profiler pipeline. A profiler run is still one process today, but the controller now reads meaning out of what that process prints, and the machinery to run several stages in order is in place and tested behind the existing single-stage API.

### Reading the tool's output

`ProfilerScrapeEngine` matches a controller-owned rule table against the child's console as it arrives. Matching is per complete line: partial output is held until a `\r` or `\n` arrives, so a rule cannot match across a boundary that the tool has not finished writing, and ANSI escapes are stripped first so a coloured path is still a path. A line over 4 KiB is skipped rather than buffered without limit, and the run reports how many lines it had to skip when a key goes unmatched, so a missing value is never silently attributed to the tool.

Rules live in `rocprofvis_controller_profiler_scrape_rules.cpp`, keyed by tool, operation and version, with a fallback to the base rules for an unrecognised version. Keeping them in the controller rather than in each backend means the value a run produces does not depend on which caller launched it.

A scraped value carries a status rather than just a string: `Pending` while the owning stage may still produce it, then `Resolved`, `Unmatched`, `RuleFailed` if the rule itself threw and was disabled, or `StageSkipped` if the stage never ran. Slots are keyed by stage as well as name, so the same tool in two stages produces two independent values.

The rocprof-sys trace rule is split into `Database:`, then `File:`, then a bare `.db` sweep, so rule order expresses the same preference the View's `ParseTraceOutputPath` documents. Both readers of one log now pick the same file.

### Running stages in order

A config can hold several stages. Each advances on a zero exit, gets its own executor, and can name the previous stages' scraped values as `{stageN.key}` placeholders in its argv, working directory and environment. Tools are resolved once during pre-flight, so a stage is never resolved twice and a missing tool fails the run before anything spawns.

A config with no stages runs its flat tool, argv, environment and working directory as a single stage. That is what every caller does today, so the existing path is the one-element case rather than a parallel implementation, and remote launches go through the same bookkeeping so the stage getters behave identically either way.

The controller exposes stage count, per-stage state, the index of the failing stage, and the pipeline's artifact path. The failing index matters for a later UI affordance: when capture succeeds and analyze fails, only the cheap stage needs re-running, and the UI can only offer that if it knows which one broke.

### Cancel, the exit code, and the output drain

`Cancel` previously marked slots skipped and returned, leaving `UpdateState` to finalise a run it no longer considers running, so `EndStage` never ran: a scraped relative path stayed relative and the last unterminated line was never matched. Cancel now drains and ends the stage itself, and the engine ignores anything fed outside a `BeginStage`/`EndStage` pair, so the drain that follows the executor cannot resolve a slot just reported as abandoned.

**This makes cancel semantics explicit, so say so if you disagree:** a cancelled stage keeps what it scraped, because a capture killed part way through has still written the file it named. Only stages that never start are marked skipped.

The kill happens with `m_mutex` released. The executor gives the child a SIGTERM grace period and can then block in `waitpid`, which a process parked in a driver call does not leave promptly, and `GetOutput` takes the same lock — so holding it would freeze the console the user is watching. A flag set under the lock makes Cancel the sole owner of the ending, so no stage is settled twice and a boundary already in flight cannot start the next one.

`GetExitCode` is written when a stage ends and read from the UI thread through the C ABI, so it takes the lock now. Cancel sets it too, rather than leaving a cancelled second stage reporting the first one's code.

On Windows `ReadOutput` did one 4 KiB read per call, so bytes still in the pipe when the child exited were never read and a chatty tool's console could truncate; it drains until the pipe is empty now. Each executor also kept a second unbounded copy of the whole run's output that nothing ever read.

### Instrument mode emits a valid command line

`rocprof-sys-instrument` treats `-o`/`--output` as the filename for a rewritten binary and then exits without running the app, and `--preset`/`--trace` are registered only on `run` and `sample`, so passing them to `instrument` is a parse error. `FlattenToExecution` now dispatches to `FlattenRunOrSample` or `FlattenInstrumentRuntime`: one-shot runtime instrumentation carries the output folder on `ROCPROFSYS_OUTPUT_PATH` only and enables Perfetto through the environment instead of the command line.

### Cancel's result reaches the caller

`rocprofvis_profiler_cancel` returned success unconditionally, discarding the controller's answer. It propagates now, and `ProfilerSessionBase::Cancel` returns `rocprofvis_result_t` instead of `bool`, so the orchestrator can tell "stopped it" from "the run finished between the click and the call". The latter previously showed as a spurious "Failed to cancel profiler".

## Tests

371 assertions in 53 cases on Linux, 223 in 34 on Windows. New coverage for line assembly and the 4 KiB cap, rule versioning and fallback, the shipped rocprof-sys preference order, output arriving after a stage ends, placeholder resolution, stage advancement and skipping, and cancel preserving a stage's scraped values while resolving relative paths against that stage's working directory.

This also fixes a test that has been failing on Linux in main: `A missing working directory fails the run instead of running elsewhere` still expected child exit 126 and `chdir failed`, but the check moved ahead of the spawn, so it is a launch error with the reason on the console. Windows CI never caught it because the test is POSIX-only.

`rocprofvis_controller_profiler_pipeline_tests.cpp` is added but deliberately not in the build: it drives the stage C ABI, which does not exist until Phase 2. The loop it would exercise is covered meanwhile by `rocprofvis_controller_profiler_pipeline_loop_tests.cpp`, which uses the controller's C++ classes directly.

## Notes for reviewers

Nothing in the View drives a multi-stage run yet. The C ABI gains the stage handle typedef and the operation and scrape-status enums, but not the stage functions, so this cannot change the behaviour of any launch the UI can currently start.

One latent issue is left for Phase 2 rather than fixed here: `UpdateState` still holds `m_mutex` across artifact relocation at a stage boundary, which could stall `GetOutput` if a relocation ever moves a large file. It is unreachable today, since relocation only runs for a multi-stage config and nothing can build one.


src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    if (outcome == CancelOutcome::kNotRunning)
     {
-        m_state = kRPVProfilerStateCancelled;
-        return kRocProfVisResultSuccess;
+        // The child had already exited. Hand the ending back to UpdateState,
+        // which settles the stage on the status the executor collected.
+        m_cancel_requested = false;
```
drchen-amd: **Cancel racing stage completion can restart the pipeline**

Clearing `m_cancel_requested` here hands the ending back to `UpdateState`, which calls `FinishStageLocked`. On `exit_code == 0` that unconditionally calls `StartStageLocked(m_current_stage + 1)` (:1394).

So if stage N finishes successfully in the window between the user hitting cancel and `executor->Cancel()` being reached, stage N+1 launches *after* the cancel. The header comment on `m_cancel_requested` (rocprofvis_controller_profiler_process.h:281-287) says the flag is never cleared while a run is ending, precisely to prevent this — so I may be misreading an invariant held elsewhere. Is there something on this path that stops the advance?

Separately: returning `kRocProfVisResultUnknownError` for "the child had already exited" gives callers the same code as the `kRefused` branch below (:1270). That erases the distinction `CancelOutcome` was introduced for in 9e5b8c9. The early-out at :1208 already uses `kRocProfVisResultNotSupported` for not-running, which looks like the natural code here.

<sub>🤖 AI-generated review comment</sub>

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+    m_stage_states[m_current_stage] = kRPVProfilerStateCompleted;
+    RelocateArtifactLocked(m_current_stage);
+
+    if (m_current_stage + 1 >= m_stages.size())
+    {
+        m_state = kRPVProfilerStateCompleted;
+        spdlog::info("Profiler completed successfully");
+        return;
+    }
+
```
drchen-amd: **A successful stage's exit code survives a failed stage start**

`m_exit_code` was set to this stage's code (0) at :1371. If `StartStageLocked(m_current_stage + 1)` fails here, `m_state` becomes `Failed` but `m_exit_code` stays 0 — so `rocprofvis_profiler_get_exit_code` reports success for a run that failed.

This is the hazard `EXIT_CODE_NO_STATUS` exists to prevent (comment at :37-46). The same applies to the two early-failure paths inside `StartStageLocked` — working directory invalid (:976) and `Start()` failure (:1026) — which leave the previous stage's code in place.

Suggest resetting `m_exit_code = EXIT_CODE_NO_STATUS` on every path where no process actually ran.

<sub>🤖 AI-generated review comment</sub>

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -383,70 +448,79 @@ bool LocalProfilerExecutor::IsRunning()
...
-bool LocalProfilerExecutor::Cancel()
+CancelOutcome LocalProfilerExecutor::Cancel()
 {
     if (m_process_handle == nullptr)
     {
-        return false;
+        return CancelOutcome::kNotRunning;
     }
 
     if (TerminateProcess(m_process_handle, 1))
```
drchen-amd: **Windows cancel fabricates exit code 1 and does not wait for the process**

Two separate things:

**1. The exit code is invented.** POSIX reports `128 + SIGKILL` = 137 via `set_exit_code_from_status` (:553). This reports 1, indistinguishable from a tool that genuinely failed with status 1. The `kRefused` branch 15 lines down explicitly declines to "borrow a number that would read as a real result" — this branch does exactly that borrowing. Passing the sentinel as `TerminateProcess`'s `uExitCode` argument would make `GetExitCodeProcess` report it consistently and fix this for free.

**2. `TerminateProcess` is asynchronous.** It returns once termination is *requested*, not complete. The POSIX path does SIGTERM → grace → SIGKILL → blocking `waitpid` (:749-770); here we set `m_is_running = false` and return `kStopped` immediately, so `RelocateArtifactLocked` can try to rename an artifact the dying child still holds open — a sharing violation on Windows. A `WaitForSingleObject` before returning would close this.

<sub>🤖 AI-generated review comment</sub>

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+    if (outcome == CancelOutcome::kRefused)
+    {
+        // The process is still running and cannot be stopped. Settling the run
+        // here is the only honest end: handing it back to UpdateState would
+        // report Running for the rest of that process's life, and the monitor
+        // job no longer waits on an executor it cannot end. m_cancel_requested
+        // stays set - the ending is not going back to UpdateState, and a second
+        // Cancel on a finished run should say NotSupported.
+        DrainExecutorLocked();
+        m_scrape.EndStage(m_stages[m_current_stage].working_directory);
```
drchen-amd: **Unchecked index sits directly above the bounds check meant to guard it**

`m_stages[m_current_stage]` is indexed here, and the very next statement (:1257) guards the same index with `if (m_current_stage < m_stage_states.size())`. If the index can be out of range, the UB happens on this line before the guard runs; if it cannot, the guard is dead code.

`m_stages` and `m_stage_states` are sized together in `PreparePipeline` (:891) so they agree today — this is about the two adjacent lines disagreeing on whether that invariant is trusted. Same pattern at :1279-1283 and :1373.

<sub>🤖 AI-generated review comment</sub>

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -746,72 +842,246 @@ void ProfilerProcessController::EndMonitorJob()
...
+    // stage is what lets the caller say which one is misconfigured.
+    for (size_t i = 0; i < m_stages.size(); ++i)
     {
-        m_state = kRPVProfilerStateFailed;
-        return resolved;
+        rocprofvis_result_t resolved = ProfilerTool::ResolvePath(
+            m_stages[i].tool, m_stages[i].tool_directory, m_stage_tool_paths[i]);
+        if (resolved != kRocProfVisResultSuccess)
+        {
+            m_failing_stage = static_cast<int32_t>(i);
```
drchen-amd: **Tool-resolution failure leaves the failing stage reported as `Idle`**

`m_failing_stage = i` is set here, but `m_stage_states[i]` is never moved off `kRPVProfilerStateIdle`, so `GetStageState(i)` reports not-yet-started for the very stage `GetFailingStage()` names as the failure. A UI highlighting the failing stage would show it as pending.

The three other failure paths (:964, :981, :1032) all set `kRPVProfilerStateFailed` alongside `m_failing_stage`. Since `m_stage_states` is already sized and Idle-filled above, this is a one-line addition.

<sub>🤖 AI-generated review comment</sub>

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+
+    if (outcome == CancelOutcome::kNotRunning)
     {
-        m_state = kRPVProfilerStateCancelled;
-        return kRocProfVisResultSuccess;
+        // The child had already exited. Hand the ending back to UpdateState,
+        // which settles the stage on the status the executor collected.
+        m_cancel_requested = false;
+        return kRocProfVisResultUnknownError;
     }
```
drchen-amd, line 1244: **Cancel racing stage completion can restart the pipeline**

Clearing `m_cancel_requested` here hands the ending back to `UpdateState`, which calls `FinishStageLocked`. On `exit_code == 0` that unconditionally calls `StartStageLocked(m_current_stage + 1)` (:1394).

So if stage N finishes successfully in the window between the user hitting cancel and `executor->Cancel()` being reached, stage N+1 launches *after* the cancel. The header comment on `m_cancel_requested` (rocprofvis_controller_profiler_process.h:281-287) says the flag is never cleared while a run is ending, precisely to prevent this — so I may be misreading an invariant held elsewhere. Is there something on this path that stops the advance?

Separately: returning `kRocProfVisResultUnknownError` for "the child had already exited" gives callers the same code as the `kRefused` branch below (:1270). That erases the distinction `CancelOutcome` was introduced for in 9e5b8c9. The early-out at :1208 already uses `kRocProfVisResultNotSupported` for not-running, which looks like the natural code here.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+    {
+        m_state = kRPVProfilerStateCompleted;
+        spdlog::info("Profiler completed successfully");
+        return;
+    }
+
+    if (StartStageLocked(m_current_stage + 1) != kRocProfVisResultSuccess)
+    {
+        m_state = kRPVProfilerStateFailed;
+    }
```
drchen-amd, line 1398: **A successful stage's exit code survives a failed stage start**

`m_exit_code` was set to this stage's code (0) at :1371. If `StartStageLocked(m_current_stage + 1)` fails here, `m_state` becomes `Failed` but `m_exit_code` stays 0 — so `rocprofvis_profiler_get_exit_code` reports success for a run that failed.

This is the hazard `EXIT_CODE_NO_STATUS` exists to prevent (comment at :37-46). The same applies to the two early-failure paths inside `StartStageLocked` — working directory invalid (:976) and `Start()` failure (:1026) — which leave the previous stage's code in place.

Suggest resetting `m_exit_code = EXIT_CODE_NO_STATUS` on every path where no process actually ran.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -383,70 +448,79 @@ bool LocalProfilerExecutor::IsRunning()
...
+        return CancelOutcome::kNotRunning;
     }
 
     if (TerminateProcess(m_process_handle, 1))
     {
         m_is_running = false;
         m_exit_code = 1;
-        return true;
+        return CancelOutcome::kStopped;
     }
```
drchen-amd, line 463: **Windows cancel fabricates exit code 1 and does not wait for the process**

Two separate things:

**1. The exit code is invented.** POSIX reports `128 + SIGKILL` = 137 via `set_exit_code_from_status` (:553). This reports 1, indistinguishable from a tool that genuinely failed with status 1. The `kRefused` branch 15 lines down explicitly declines to "borrow a number that would read as a real result" — this branch does exactly that borrowing. Passing the sentinel as `TerminateProcess`'s `uExitCode` argument would make `GetExitCodeProcess` report it consistently and fix this for free.

**2. `TerminateProcess` is asynchronous.** It returns once termination is *requested*, not complete. The POSIX path does SIGTERM → grace → SIGKILL → blocking `waitpid` (:749-770); here we set `m_is_running = false` and return `kStopped` immediately, so `RelocateArtifactLocked` can try to rename an artifact the dying child still holds open — a sharing violation on Windows. A `WaitForSingleObject` before returning would close this.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1193,314 @@ void ProfilerProcessController::ClearOutput()
...
+        // The process is still running and cannot be stopped. Settling the run
+        // here is the only honest end: handing it back to UpdateState would
+        // report Running for the rest of that process's life, and the monitor
+        // job no longer waits on an executor it cannot end. m_cancel_requested
+        // stays set - the ending is not going back to UpdateState, and a second
+        // Cancel on a finished run should say NotSupported.
+        DrainExecutorLocked();
+        m_scrape.EndStage(m_stages[m_current_stage].working_directory);
+
+        if (m_current_stage < m_stage_states.size())
```
drchen-amd, line 1257: **Unchecked index sits directly above the bounds check meant to guard it**

`m_stages[m_current_stage]` is indexed here, and the very next statement (:1257) guards the same index with `if (m_current_stage < m_stage_states.size())`. If the index can be out of range, the UB happens on this line before the guard runs; if it cannot, the guard is dead code.

`m_stages` and `m_stage_states` are sized together in `PreparePipeline` (:891) so they agree today — this is about the two adjacent lines disagreeing on whether that invariant is trusted. Same pattern at :1279-1283 and :1373.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -746,72 +842,246 @@ void ProfilerProcessController::EndMonitorJob()
...
-        return resolved;
+        rocprofvis_result_t resolved = ProfilerTool::ResolvePath(
+            m_stages[i].tool, m_stages[i].tool_directory, m_stage_tool_paths[i]);
+        if (resolved != kRocProfVisResultSuccess)
+        {
+            m_failing_stage = static_cast<int32_t>(i);
+            m_scrape.SkipRemainingFrom(0);
+            spdlog::error("Profiler stage {} tool could not be resolved", i);
+            return resolved;
+        }
```
drchen-amd, line 921: **Tool-resolution failure leaves the failing stage reported as `Idle`**

`m_failing_stage = i` is set here, but `m_stage_states[i]` is never moved off `kRPVProfilerStateIdle`, so `GetStageState(i)` reports not-yet-started for the very stage `GetFailingStage()` names as the failure. A UI highlighting the failing stage would show it as pending.

The three other failure paths (:964, :981, :1032) all set `kRPVProfilerStateFailed` alongside `m_failing_stage`. Since `m_stage_states` is already sized and Idle-filled above, this is a one-line addition.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
tomk-amd: Windows implementation is present - but realistically out-of-scope and not tested.  No rocm profilers currently   work on the windows platform.  This can get proper attention once there is something to test against.

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
tomk-amd: This was deferred - as the current implementation phase does not actually execute multiple stages yet.

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
tomk-amd: Fixes for the AI review observations, in this commit:  https://github.com/ROCm/roc-optiq/pull/1038/commits/a4c50c4ede98dd8b8c3e061bfa2bedd2bc502458

See commit comments for details.

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -383,70 +448,106 @@ bool LocalProfilerExecutor::IsRunning()
...
-                if (ReadFile(m_stderr_read_handle, buffer, sizeof(buffer) - 1, &bytes_read, nullptr))
-                {
-                    buffer[bytes_read] = '\0';
-                    output += buffer;
-                }
-            }
+            return;
         }
+        output.append(buffer, bytes_read);
     }
```
drchen-amd: **`m_output_text` grows without bound, and `m_mutex` is held across the whole drain**

`read_pipe_available` loops until the pipe reports empty, and every byte it returns is appended to `m_output_text` (:1209) and never trimmed. A long run on a chatty workload — an ML training job whose stdout this stream also carries — accumulates the entire session in memory, and `GetOutput` returns a full copy of it on every 100 ms poll.

Secondary: `GetOutput` (:1221-1223) holds `m_mutex` for the duration of the drain, so `GetState`, `GetExitCode`, `GetStageState` and `Cancel`'s first critical section all block behind it. In practice the loop drains faster than a child produces, so this is latency rather than a stall — but a byte cap per call would bound both, and the 100 ms poll would pick up the remainder on the next tick.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.h
```diff
@@ -145,8 +175,10 @@ class LocalProfilerExecutor : public IProfilerExecutor
 
     std::atomic<bool> m_is_running;
     int m_exit_code;
```
drchen-amd: **`m_exit_code` is written from two threads without synchronisation**

`m_is_running` is correctly `std::atomic<bool>`, but `m_exit_code` next to it is a plain `int` written from both threads:

- the job thread, via `UpdateState` → `IsRunning()` (:443 Windows, :792 POSIX), holding the controller's `m_mutex`
- the UI thread, via `Cancel()` (:480, :485, :499) — and `ProfilerProcessController::Cancel` deliberately *releases* `m_mutex` before calling it (:1265-1272)

So the controller's lock does not serialise these two writers. `m_cancel_requested` narrows the window but does not close it: `UpdateState` can already be inside `IsRunning()` when `Cancel` sets the flag.

`SshProfilerExecutor` uses `std::atomic<int>` for the same field, so the two executors disagree on the contract.

On POSIX there is a second edge: both threads can `waitpid` the same pid, one reaps and the other gets `ECHILD`, which writes `m_exit_code = 1` (:792) over the status the first thread just recorded.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
```diff
@@ -892,48 +1231,336 @@ void ProfilerProcessController::ClearOutput()
...
+        // The child exited on its own in the window between the user asking to
+        // stop and Cancel reaching it, so Cancel handed the ending back here.
+        // This stage keeps the result it earned, but starting the next one now
+        // would run work the user has already asked to abandon.
+        m_scrape.SkipRemainingFrom(m_current_stage + 1);
+        m_state = kRPVProfilerStateCancelled;
+        spdlog::info("Profiler stage {} finished as the run was cancelled; stage {} not started",
+                     m_current_stage, m_current_stage + 1);
+        return;
+    }
```
drchen-amd: **This branch — the subtlest part of the fix — has no test**

The two cases added in a4c50c4 cover the exit-code drop and the failing-stage state, but nothing exercises this path: a child that exits on its own inside the cancel window, so `Cancel` returns `kNotRunning`, clears `m_cancel_requested`, and hands the ending back here with `m_stop_requested` still set.

Worth noting it is unreachable today for a different reason than the deferral discussion above — with a single-stage pipeline the `m_current_stage + 1 >= m_stages.size()` check at :1430 returns first, so `m_state` becomes `Completed` and this branch never runs. It only becomes live when a second stage exists, which is exactly when a regression here would be hardest to spot.

`config.AddStage` already works from the C++ API — the new test in this commit uses it — so a two-stage pipeline where stage 1 exits normally while a cancel is in flight is expressible today, without waiting for the Phase 2 C ABI.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
drchen-amd, line 541: **The drain loop has no per-call bound**

`read_pipe_available` reads until the pipe reports empty, and `GetOutput` (:1221-1223) holds `m_mutex` for the whole call, so `GetState`, `GetExitCode`, `GetStageState` and `Cancel`'s first critical section all wait behind it. In practice the loop drains faster than a child produces, so this is added latency rather than a stall — but the bound is set by how fast the child writes, not by anything here.

A byte cap per call (1 MiB, say) would bound the hold; `ExecuteJob`'s 100 ms poll picks up the remainder on the next tick.

Related but pre-existing, so just noting it rather than asking for a change here: `m_output_text` accumulates every byte for the life of the run with no high-water mark (main does the same at its :880), and `GetOutput` returns a full copy of it on every poll. A long run on a chatty workload — an ML job whose stdout this stream also carries — grows it without limit.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
drchen-amd: **This branch — the subtlest part of the fix — has no test**

The two cases added in a4c50c4 cover the exit-code drop and the failing-stage state, but nothing exercises this path: a child that exits on its own inside the cancel window, so `Cancel` returns `kNotRunning`, clears `m_cancel_requested`, and hands the ending back here with `m_stop_requested` still set.

Worth noting it is unreachable today for a different reason than the deferral discussion above — with a single-stage pipeline the `m_current_stage + 1 >= m_stages.size()` check at :1430 returns first, so `m_state` becomes `Completed` and this branch never runs. It only becomes live when a second stage exists, which is exactly when a regression here would be hardest to spot.

`config.AddStage` already works from the C++ API — the new test in this commit uses it — so a two-stage pipeline where stage 1 exits normally while a cancel is in flight is expressible today, without waiting for the Phase 2 C ABI.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_scrape_rules.cpp
```diff
@@ -0,0 +1,266 @@
...
+// Row order is the preference order, and it is the same one the View's
+// ParseTraceOutputPath documents, so the two readers cannot pick different
+// files off the same log: the "Database:" log line first, then the Output
+// Summary box's "File:", then a bare-token sweep for a tool that labels
+// neither. Collapsing the first two into one alternation would let a later
+// summary line outrank the database line, which is the disagreement.
+constexpr RuleRow kRocprofSysTrace[] = {
+    {kKeyTraceDb, R"(Database:\s*'?([^\s'"]+\.db)\b)", 1, ScrapePolicy::kLastMatch},
+    {kKeyTraceDb, R"(File:\s*'?([^\s'"]+\.db)\b)", 1, ScrapePolicy::kLastMatch},
+    {kKeyTraceDb, R"(([^\s'"]+\.db)\b)", 1, ScrapePolicy::kLastMatch},
```
drchen-amd, line 67: **The bare-token fallback will match a `.db` inside an error message**

The precedence guard does its job here — this rule only wins when neither labelled rule matched anywhere in the stage, and the comment above explains that ordering clearly.

The concern is what happens on exactly that run. `([^\s'"]+\.db)\b` is unanchored, so it matches any whitespace-delimited token ending in `.db`. On a tool that labels neither, a line like `could not open /tmp/stale.db`, or a usage string naming a `.db` argument, becomes the reported trace artifact. `kLastMatch` means the last such mention wins — and on a failing run the last mention is more likely to be a diagnostic than a real output path.

Requiring a line-start anchor, or only applying the sweep when the stage exited 0, would keep the fallback without letting an error message supply the answer.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_scrape.cpp
```diff
@@ -0,0 +1,674 @@
...
+    {
+        return false;
+    }
+
+    uint32_t stage = 0;
+    while (i < token.size() && std::isdigit(static_cast<unsigned char>(token[i])))
+    {
+        stage = stage * 10u + static_cast<uint32_t>(token[i] - '0');
+        ++i;
+    }
```
drchen-amd, line 563: **Stage index accumulates without an overflow guard**

`stage = stage * 10u + ...` wraps silently on a `uint32_t`. `{stage99999999999.key}` parses successfully as some unrelated small index, which then either resolves against the wrong stage or reports "no stage produces that key" — a misleading message for what is really a malformed placeholder.

`ParseVersion` in `..._scrape_rules.cpp` already guards this case by accumulating into a wider type and range-checking, so the two parsers in the same subsystem disagree.

Low severity, since placeholders are authored in code rather than by users today.

_🤖 AI-generated review comment_

src/controller/CMakeLists.txt
```diff
@@ -146,7 +149,15 @@ if(BUILD_TESTING)
...
-        add_executable(${PROFILER_TEST_NAME} tests/rocprofvis_controller_profiler_tests.cpp)
+        # rocprofvis_controller_profiler_pipeline_tests.cpp is deliberately not
+        # here yet: it drives the stage C ABI, which does not exist until
+        # Phase 2. The loop it would exercise is covered meanwhile by
+        # rocprofvis_controller_profiler_pipeline_loop_tests.cpp, which uses the
+        # controller's C++ classes directly.
+        add_executable(${PROFILER_TEST_NAME}
+            tests/rocprofvis_controller_profiler_tests.cpp
+            tests/rocprofvis_controller_profiler_scrape_tests.cpp
+            tests/rocprofvis_controller_profiler_pipeline_loop_tests.cpp)
```
drchen-amd: **Nothing will signal when this file should come back**

The reasoning is sound — the stage C ABI does not exist yet, and the loop tests cover the same ground through the C++ classes.

The risk is mechanical: 521 lines referencing about a dozen symbols that do not exist, excluded from the build, with no `#error`, no failing placeholder, and no ticket reference. Between now and Phase 2 it drifts against the ABI it asserts on, and whoever eventually adds it to `add_executable` meets a wall of compile errors and is most likely to delete it rather than repair it.

A ticket reference in this comment would make it actionable. Gating it behind a CMake option that Phase 2 flips would make the breakage surface at the moment it becomes fixable.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
drchen-amd: Agreed on scoping — this is fine to leave until there is a Windows profiler to test against.

One note on the a4c50c4 version, for whenever that happens: if `WaitForSingleObject` times out, the child is still alive, but the function sets `m_is_running = false` and returns `kStopped`. The controller then treats the stage as ended — `DrainExecutorLocked`, `EndStage`, and `RelocateArtifactLocked` renaming artifact files the live process still holds open, which is the sharing violation the wait was added to prevent.

`kRefused` already describes exactly that situation ("still running and cannot be stopped"), and the controller handles it. The timeout branch could return it rather than `kStopped`.

_🤖 AI-generated review comment_

src/controller/src/profiler/rocprofvis_controller_profiler_scrape.cpp
tomk-amd: this is nitpicking nonsense - currently profiling will have 2 stages, possibly a couple more if we extend this functionality to include some additional operations.

src/controller/src/profiler/rocprofvis_controller_profiler_scrape_rules.cpp
tomk-amd: In no way a blocker, at worst upon an already failed profiler execution, a non-existant database would also fail downloading.  The system will already be in an error state.

src/controller/src/profiler/rocprofvis_controller_profiler_scrape.cpp
tomk-amd: after looking a bit more deeply, i didn't realize this referred to a number "scraped" from the command line arguments.. still not an issue in my opinion, a user would have to enter a value like: {stage99999999999.trace_db} into one of the text boxes that gets appended to the command line.  The side effect is the capture session would fail in some way - which is probably the correct thing to do with a nonsense argument.

src/controller/src/profiler/rocprofvis_controller_profiler_process.cpp
tomk-amd: addressed in commit https://github.com/ROCm/roc-optiq/pull/1038/commits/bdd73f5f099bdbdc2569dc436baaad1b939f0dd1

src/view/src/rocprofvis_settings_manager.cpp
```diff
@@ -995,6 +995,9 @@ void
 SettingsManager::DeserializeProfilerSettings(jt::Json& json)
 {
     jt::Json& ps = json[JSON_KEY_GROUP_SETTINGS][JSON_KEY_SETTINGS_CATEGORY_PROFILER];
+    // A "profiler_path" key written by an older development build is read by
+    // nothing: the path is no longer part of the launch input, so a settings file
+    // cannot name the binary to execute.
```
drchen-amd: Is this referring to something that got removed?

drchen-amd (approved):

src/view/src/rocprofvis_settings_manager.cpp
tomk-amd: yes - the executable path used be stored / populated on the UI side - but in true AI fashion this comment is useless now, will remove.

PR #1040, opened 2026-09-07
Update metric fetching policy

amokiche-amd:
## Motivation
Adapt the controller and UI for files missing metrics or PC sampling data.

## Technical Details
Compute profiler — disabled tabs for metric-free and ISA-free databases
Tabs that require data the loaded database does not contain (Table View, Baseline Comparison, ISA View) are now disabled with a tooltip explaining why, instead of being active but empty. The ISA View tab is hidden unless the database contains at least one kernel with ISA lines (has_isa_lines column added to the kernel query, schema version ≥ 2.2).

Error dialog for invalid compute databases
Replaced the ad-hoc two-boolean popup mechanism (m_database_error_queued + show_popup) with a three-state ErrorDialogState enum (kNone / kPending / kShown). This makes the lifecycle explicit, prevents duplicate dialogs from competing call sites, and moves dialog consumption from Render() to Update() so the dialog fires even if rendering is skipped. The same consistency fixes were applied to TraceView (system profiler): typedef struct replaced with plain struct, popup consumption moved to Update().

m_view_created retry guard fixed in both views
ComputeView::Update() and TraceView::Update() previously set m_view_created = true unconditionally after calling CreateView(), even when CreateView() returned early due to an error or empty workloads. Both now condition the flag on whether the primary child view was actually constructed (m_tab_container != nullptr / m_timeline_view != nullptr), allowing a retry on the next state transition.

Tab container — signed/unsigned comparison fix
TabContainer::Render() compared new_selected_tab (int, sentinel value -1) against m_tabs.size() (size_t), which promoted -1 to SIZE_MAX and silently suppressed the kTabSelected event whenever all tabs were disabled. Fixed to match the existing is_enabled_index pattern: new_selected_tab >= 0 && new_selected_tab < static_cast<int>(m_tabs.size()).

Baseline Comparison — stale results cleared on metric-free workload selection
When a workload with no metrics is selected for either the baseline or target side, the comparison view previously left the previous kernel's data visible. FetchMetrics() now explicitly clears the metric store and triggers a data-changed event for that side. UpdateMetrics() now clears m_categories and resets m_tab_container if either store is absent, ensuring stale results are never displayed as current data.

Database — kernel-free workloads now included in workload lookup set
m_workload_id_set was populated only from compute_kernel rows, so a workload with no kernels was absent from the set and caused GetComputeMetricValuesByWorkload() to return InvalidParameter instead of a valid empty result. A new query (SELECT workload_id FROM compute_workload) and callback (CallbackStoreWorkloadIdSet) now populate the set directly from the workload table at load time.

Test fixes
rocprofvis_view_test_access.h: ComputeViewTestPeer::PopupPending() updated to check ErrorDialogState::kPending after show_popup was removed from popup_info_t.
New controller and data-model tests covering the kernel-free workload path, ISA-line detection, and disabled-tab state.
image
image
image

## Submission Checklist


tomk-amd:
One minor item.  Disabled tabs will still handle events that they subscribe to when created. And those handlers may cause a state change within the class the tab holds, that then performs some kind of functionality or executes on the Update() call, which is also called when tab is disabled..  For example, the ISA view tab tries to fetch when kernel selection changes, you can see the failures in the log (might need to increase the log viewer's line count from 512 via settings to something larger to see these)

image
 
 For this specific case this is relatively benign, just a warning because data failed to fetch, but we may need to come up with a strategy to handle this more elegantly.

My first instinct is to populate the disabled tabs with a dummy object (ComputeEmptyView) instead of ComputeIsaView for example when no pc sampling data is present, but that would not solve the case where multiple workloads are present and some have ISA info while others don't.  Also not sure if all the information needed to make the choice is present at the time of ComputeView::CreateView() where the tabs are populated...

This something we should look into but can be done in separate PR.

amokiche-amd:
> One minor item. Disabled tabs will still handle events that they subscribe to when created. And those handlers may cause a state change within the class the tab holds, that then performs some kind of functionality or executes on the Update() call, which is also called when tab is disabled.. For example, the ISA view tab tries to fetch when kernel selection changes, you can see the failures in the log (might need to increase the log viewer's line count from 512 via settings to something larger to see these)
> 
> image
> For this specific case this is relatively benign, just a warning because data failed to fetch, but we may need to come up with a strategy to handle this more elegantly.
> 
> My first instinct is to populate the disabled tabs with a dummy object (ComputeEmptyView) instead of ComputeIsaView for example when no pc sampling data is present, but that would not solve the case where multiple workloads are present and some have ISA info while others don't. Also not sure if all the information needed to make the choice is present at the time of ComputeView::CreateView() where the tabs are populated...
> 
> This something we should look into but can be done in separate PR.

Agreed, that’s a good point; I’ll get to work on it.

tomk-amd:
small suggestion, freeze the header row in the table(s):

image

image

tomk-amd (approved):

PR #1046, opened 2026-09-09
Bind metric definitions to their real entry id (fixes >100% memory chart values)

feizheng10:
## Problem

The memory chart shows cache hit rates well above 100% and disagrees with the `rocprof-compute` CLI for the same kernel. Reported as AIPROFCOMP-842.

For `__amd_rocclr_fillBufferUnAligned` in an MI300A profile (analysis schema 2.2.0), the chart rendered:

| Row | Chart (before) | Database / CLI |
| --- | --- | --- |
| `VL1 Hit` | 0.00 % | 0 % |
| `sL1D Hit` | **4.19M %** | 0 % |
| `IL1 Hit` | **360 %** | 100 % |
| `L2 Hit` | **2.05B %** | 50 % |
| `VL1 Coalescing` | 0.00 % | 25 % |
| `L2 Rd` / `L2 Wr` / `L2 Atomic` | 530K / 765 / 2.28K | 2798 / 4194300 / 0 |
| `LDS Latency` | 262K cycles | NULL (should be N/A) |

The generated database is correct: all 67 memory-chart metrics for that kernel are in range. The corruption is entirely on the read path.

## Root cause

Metric values are keyed by the trailing component of their `category.table.entry` metric id, parsed from the `metric_id` string in `DataProvider::OnMetricsFetched`. The definitions that supply the *names* were keyed by row position instead:

```cpp
// Last position of id is not returned, for now assume index...
workload.available_metrics.list[j].id = static_cast<uint32_t>(table.entries.size());
table.entries.insert({ static_cast<uint32_t>(table.entries.size()), ... });
```

That assumption does not hold. `compute_metric_definition` declares `UNIQUE (workload_id, metric_id)`, so `WHERE workload_id = ?` is served from that autoindex and SQLite returns rows ordered by `metric_id` **as a string** — `3.1.10` before `3.1.2`. Any table with eleven or more metrics therefore binds each name to whichever value sits at its *lexicographic* position.

Instrumenting `LoadMetricList` confirms it directly: definition `j=2` is `VGPR` (`3.1.10`) while value `entry=2` is `SALU` (`3.1.2`).

The shift explains every wrong cell. `L2 Hit` is `3.1.52`, lexicographic position 48, and the 2.05B it displayed is exactly the value of `3.1.48 IL1_L2 Read BW` — a bandwidth counter rendered with a `%` suffix.

This is also why #901 (switching the memory chart from positional to name-based lookup) did not fix the symptom: the name-based lookup was correct, but the names themselves were attached to the wrong values one layer down. The bug is present in `v1.0.0-optiq` and on current `main`, and it affects the repo's own `sample/rocprof_compute_23ed6f36.db` (55 memory-chart metrics) too.

## Fix

Expose the trailing component of `metric_id` as an `entry_id` column on the metrics-definition query, parsed in the same callback that already splits out the category and table components, and carry it through `kRPVComputeColumnEntryId` and `kRPVControllerWorkloadAvailableMetricEntryIdIndexed` into `LoadMetricList`. Definitions and values then agree on the key.

The metric definition contract requires a complete `category.table.entry` id; `LoadMetricList` now requires the controller entry-id lookup to succeed and uses that id directly.

## Verification

Built from `main` and opened the reproduction database. Every memory-chart cell now matches the database exactly, including the four hit rates above, `VL1 Stall` 67 cycles, `sL1D Latency` 58.3K cycles, `IL1 Latency` 765 cycles, `Fabric Rd Lat` 1.11K cycles, and `HBM Wr` 4.19M. `LDS Latency` and `Fabric Atomic Lat` correctly report `N/A` where the database stores NULL, instead of borrowing a neighbour's value.

Also verified against `sample/rocprof_compute_23ed6f36.db`, which was affected too: its first kernel now shows `IL1 Hit` 100 %, `L2 Hit` 84 %, `sL1D Hit` 88 %, `VL1 Coalesce` 25 %, `IL1 Lat` 725 — matching the stored values.

Tests:

- `datamodel-compute-tests` — all passed (472,842 assertions)
- `roc-optiq-controller-compute-tests` — all passed (1,015,245 assertions)

The two system-test suites fail both with and without this change, because `sample/trace_70b_1024_32.rpd` is a Git LFS pointer rather than a real trace in a plain clone.

## Note on ordering

`Table::ordered_entries` sorts by entry id, so metric tables now list metrics in their config-defined numeric order rather than the string order (`3.1.10` before `3.1.2`) they used previously.

### DIFF
--- a/src/controller/src/compute/rocprofvis_controller_workload.h
+++ b/src/controller/src/compute/rocprofvis_controller_workload.h
@@ -42,13 +42,19 @@ private:
         std::vector<std::string> values;
     };
     struct MetricDefinition {
-        uint32_t category_id;
-        uint32_t table_id;
-        size_t category_name_idx;
-        size_t table_name_idx;
-        size_t name_idx;
-        size_t description_idx;
-        size_t unit_idx;
+        static constexpr uint32_t kUnknownEntryId = UINT32_MAX;
+
+        uint32_t category_id = 0;
+        uint32_t table_id = 0;
+        // Trailing component of the "category.table.entry" metric id. Left as
+        // kUnknownEntryId when the database predates the column, so consumers can
+        // fall back to positional numbering.
+        uint32_t entry_id = kUnknownEntryId;
+        size_t category_name_idx = 0;
+        size_t table_name_idx = 0;
+        size_t name_idx = 0;
+        size_t description_idx = 0;
+        size_t unit_idx = 0;
     };
     struct MetricValueName {
         uint32_t category_id;
--- a/src/controller/src/compute/rocprofvis_controller_workload.h
+++ b/src/controller/src/compute/rocprofvis_controller_workload.h
@@ -42,14 +42,9 @@ private:
         std::vector<std::string> values;
     };
     struct MetricDefinition {
-        static constexpr uint32_t kUnknownEntryId = UINT32_MAX;
-
         uint32_t category_id = 0;
         uint32_t table_id = 0;
-        // Trailing component of the "category.table.entry" metric id. Left as
-        // kUnknownEntryId when the database predates the column, so consumers can
-        // fall back to positional numbering.
-        uint32_t entry_id = kUnknownEntryId;
+        uint32_t entry_id = 0;
         size_t category_name_idx = 0;
         size_t table_name_idx = 0;
         size_t name_idx = 0;
--- a/src/view/src/rocprofvis_data_provider.cpp
+++ b/src/view/src/rocprofvis_data_provider.cpp
@@ -4712,20 +4712,11 @@ DataProvider::LoadMetricList(WorkloadInfo& workload, rocprofvis_handle_t* worklo
         table.id   = workload.available_metrics.list[j].table_id;
         table.name = GetString(workload_handle,
                                kRPVControllerWorkloadAvailableMetricTableNameIndexed, j);
-        // Metric values are keyed by the trailing component of their
-        // "category.table.entry" id, so the definitions have to use that same id.
-        // Row position is not a valid substitute: the definition query is served
-        // from the UNIQUE(workload_id, metric_id) index, which orders rows by
-        // metric_id as a string ("3.1.10" before "3.1.2"). Databases that predate
-        // the column report no entry id, so fall back to positional numbering.
-        uint32_t entry_id = static_cast<uint32_t>(table.entries.size());
-        if(rocprofvis_controller_get_uint64(
-               workload_handle, kRPVControllerWorkloadAvailableMetricEntryIdIndexed, j,
-               &uint64_data) == kRocProfVisResultSuccess &&
-           uint64_data != UINT32_MAX)
-        {
-            entry_id = static_cast<uint32_t>(uint64_data);
-        }
+        result = rocprofvis_controller_get_uint64(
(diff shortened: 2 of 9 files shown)


src/controller/src/compute/rocprofvis_controller_workload.h
```diff
@@ -42,13 +42,19 @@ class Workload : public Handle
...
-        size_t description_idx;
-        size_t unit_idx;
+        static constexpr uint32_t kUnknownEntryId = UINT32_MAX;
+
+        uint32_t category_id = 0;
+        uint32_t table_id = 0;
+        // Trailing component of the "category.table.entry" metric id. Left as
+        // kUnknownEntryId when the database predates the column, so consumers can
+        // fall back to positional numbering.
+        uint32_t entry_id = kUnknownEntryId;
```
drchen-amd: I don't think this fallback is needed and entry can just have initial 0 value like others. 

Based on callback parser, fallback will only trigger in scenario where metric_id column does not exist or metric_id only has 2/3 components. If this happens, I expect many other things to break, and fallback would not accomplish much.

src/view/src/rocprofvis_data_provider.cpp
```diff
@@ -4712,11 +4712,22 @@ DataProvider::LoadMetricList(WorkloadInfo& workload, rocprofvis_handle_t* worklo
...
+        // "category.table.entry" id, so the definitions have to use that same id.
+        // Row position is not a valid substitute: the definition query is served
+        // from the UNIQUE(workload_id, metric_id) index, which orders rows by
+        // metric_id as a string ("3.1.10" before "3.1.2"). Databases that predate
+        // the column report no entry id, so fall back to positional numbering.
+        uint32_t entry_id = static_cast<uint32_t>(table.entries.size());
+        if(rocprofvis_controller_get_uint64(
+               workload_handle, kRPVControllerWorkloadAvailableMetricEntryIdIndexed, j,
+               &uint64_data) == kRocProfVisResultSuccess &&
+           uint64_data != UINT32_MAX)
```
drchen-amd: Same reasoning as other comment, I think success return code is enough and check for fallback is not needed.

drchen-amd (commented):
Makes sense and looks good.

I did double check the db in our samples folder and could not reproduce any of the described issues. The memory chart mapped correctly to the db and the tables had the metric listed in ascending order. This is because the old ordering assumption held as long as schema had no uniqueness constraints. It looks like existing tables introduced this with schema 2.0.0 (rocm-systems #8334) and we haven't updated our side besides the pc sampling.

src/controller/src/compute/rocprofvis_controller_workload.h
feizheng10: Agreed. Removed the sentinel and now initialize `entry_id` to 0 like the other fields. A malformed or missing composite `metric_id` is already outside the supported database contract, so the fallback did not provide meaningful compatibility. Addressed in 8d6bca2f.

src/view/src/rocprofvis_data_provider.cpp
feizheng10: Agreed. `LoadMetricList` now requires a successful `kRPVControllerWorkloadAvailableMetricEntryIdIndexed` lookup and directly uses the returned id; the positional and `UINT32_MAX` fallbacks are removed. Addressed in 8d6bca2f.

drchen-amd (approved):

PR #1051, opened 2026-09-14
Add copy and export for the Ask Optiq conversation

dhingora-amd:
## Motivation

The Ask Optiq assistant had no way to get a conversation out of the panel.
This adds Copy and Export so users can save an investigation, share findings,
or paste a transcript elsewhere.

## Technical Details

- Adds **Copy** and **Export** icon buttons to the assistant composer, beside
  Clear; both stay disabled until there is a conversation to save.
- **Copy** places the transcript on the clipboard (`ImGui::SetClipboardText`).
- **Export** writes it to a Markdown/text file via `AppWindow::ShowSaveFileDialog`.
- New `BuildTranscriptText()` renders the transcript as Markdown (You / Optiq
  turns, notices as blockquotes); charts are noted rather than snapshotted
  since they redraw from live model data.
- Reuses `NotificationManager` for success/failure feedback — no new logging
  or file-write helpers introduced.
- Files: `src/view/src/agenticprofiling/rocprofvis_ai_assistant.{h,_render.cpp}`.


tomk-amd (approved):

tomk-amd:
None of these are blockers, but recording them here...

### Issues 

- **Medium (UX)** — `rocprofvis_ai_assistant_render.cpp` (Export `IconButton`):
  Export uses `ICON_ARROW_IN_BOX`, which is already the **import** icon (`"Import from Basic"` in `rocprofvis_multi_track_table.cpp`). Tooltip says export, but the glyph means the opposite. Prefer `ICON_DOCUMENT`, `ICON_ARCHIVE`, or `ICON_ARROW_FORWARD` unless a dedicated export glyph is added.
- **Low** — `ExportTranscript()` write path:
  After `is_open()`, success is toasted without checking `write`/`flush`. A full disk or later write error still looks like success. Same gap exists in `ScriptEditor::WriteFile`; worth fixing here if you touch this path.


### Nits 

- Copy/export I/O lives in `_render.cpp`, whose file comment says that file only draws ImGui and that turn/state machinery stays in `rocprofvis_ai_assistant.cpp`. Functionally fine (`assistant.cpp` already includes `AppWindow`); a later split could move these three methods.
- Native save dialogs append **the first filter’s** extension when the name has none (`AppWindow` around the NFD path). Markdown is first, so a “Text” save named `notes` becomes `notes.md`. Pre-existing dialog behavior; not introduced here.
- Header comment says `BuildTranscriptText()` “renders” the transcript; it serializes it.

Sajeeth-Wimalasuriyan (approved):

PR #1053, opened 2026-09-15
App: prefer X11/XWayland on Linux to stop GLFW Wayland error spam

dhingora-amd:
## Motivation

On native Wayland sessions the app logs `GLFW Error 65548`
(`GLFW_FEATURE_UNAVAILABLE`) every frame, flooding the log viewer
(6000+ errors in seconds). Wayland forbids querying/setting window
position, which GLFW needs for fullscreen save/restore and viewports.

## Technical Details

- Prefer the X11 platform (native or via XWayland) on Linux, only when
  an X server is reachable (`glfwPlatformSupported(GLFW_PLATFORM_X11)`
  and `DISPLAY` set); otherwise fall back to GLFW's default so
  pure-Wayland sessions still start.
- Previously gated behind `ROCPROFVIS_MULTI_WINDOW` (OFF); now applied
  to all Linux builds and mirrored in the test harness.
- Windows/macOS unaffected (guarded by `#ifdef __linux__`).


tomk-amd (approved):

tomk-amd (changes requested):
Use other pr: 1052

